PR #61, opened 2020-06-12
global: theme overwrite setup

### LINKED ISSUE #46: Make theme overridable
One possible solution is to change the Semantic-UI theme in `react-invenio-app-ils` and have it as `packaged` theme.
Since Semantic-UI would normally require to have `package` theme in their `semantic-less` repo, we will have to copy the `config` file from that module and adapt it to our needs.

## Semantic-UI limitation

If you check the `semantic-less` node module, you will find that the`Craco Less` plugin is adding a Webpack alias `"../../theme.config$" = path.join(context.paths.appSrc, "/semantic-ui/theme.config")`. The path to the `theme.config` is hardcoded.
That file contains the following:
```
@themesFolder : 'themes';
@siteFolder  : '../../src/semantic-ui/site';
@import (multiple) "~semantic-ui-less/theme.less";
```
And then the `theme.less` contains:
```
/* Packaged site.variables */
@import "@{themesFolder}/@{site}/globals/site.variables";
```
Basically the packaged theme folder is hardcoded to `node_modules/semantic-ui-less/...`.

## Solution

Copy/paste the `default` theme folder from `node_modules/semantic-ui-less/` and rename the `site` to `ils` so it becomes the packaged theme, to have:
```
semantic-ui/themes/
  default/
  ils/
```
Unfortunately the `default` is required because it is hardcoded. Adjust config files as needed.

In the overlay then, we will be able to create a site theme by changing the `theme.config`:
`@themesFolder : '~node_modules/react-invenio-app-ils/semantic-ui/themes/ils';`

kprzerwa:
* add library dist linking to package.json

closes #46


package.json
```diff
@@ -14,7 +14,13 @@
     "test": "craco test",
     "eject": "craco eject",
     "lint": "eslint src/lib/**/*.js",
-    "format": "prettier --config ./.prettierrc --ignore-path ./.prettierignore --write \"**/*.js\""
+    "format": "prettier --config ./.prettierrc --ignore-path ./.prettierignore --write \"**/*.js\"",
+    "edit-linked-package": "json -I -f ./dist/package.json -e 'this.module=\"esm/index.js\", this.main=\"cjs/index.js\"' ",
```
ntarocco, line 18: can we maybe have them prefixed by `_` to know they are kind of hack for now and these ones should be called normally by running `npm run ...`?

kprzerwa: actually I think what I did in my next PR will solve it (check craco.config.js)

PR #66, opened 2020-06-16
config: overridable configuration

### LINKED ISSUE #6: Make configuration overridable
The config files in the UI are currently hardcoded. To make it overridable, `react-invenio-app-ils` should provide a default configuration and a mechanism to override single config keys.

A possible solution would be to create a global config class, or multiple (since we have `invenioConfig`, `searchConfig`, etc...) which is instantiated ones. Then, the `App` component will have an optional prop that can receive a config object and will update the global one.

Each component or js module that needs the config will the get it from that global obj.

- create one or multiple config class with getter and setter and fill it in with the default config
- add the optional prop in App to the `set` the given config in the global one
- change all the usage of the previous config (in cmps and modules) to `get` it from the global config

**Edit** things needed after #97 
- [ ] Provide a mechanism to override `config.APP.REST_ENDOINTS_BASE_URL`
- [ ] In every api provide a way to override the exported `searchBaseURL`
- [ ] Provide a way to override the default variables in `common.js` (values that are used in our config more than once)
- [ ] convert config functions in config Class methods
- [ ] rename `invenioConfig` to `config`

topless:
- overridable configuration for the app
- closes #6


kprzerwa:
I would say we could clean up a bit, review the naming and what config variables go where but it is not blocking for this PR

src/index.js
```diff
@@ -29,9 +29,19 @@ const overriddenCmps = {
...
+    },
+  },
+  uiConfig: {},
+  searchConfig: {},
+};
+
 ReactDOM.render(
   <OverridableContext.Provider value={overriddenCmps}>
-    <InvenioILSApp />
+    <InvenioILSApp externalConfig={externalConfig} />
```
kprzerwa, line 44: nit: it could be named just `config` or `configuration`, no need for external IMO

PR #68, opened 2020-06-16
BorrowingRrequestForm: fix bug with library

### LINKED ISSUE #60: BorrowingRequestForm: fields library and item type are not accepted
## Description
The field `library` is not accepted and it displays the error `Missing data for required field`. The field `item type` is not accepted when it's value is `Physical copy`, and displays the error `Must be one of: PHYSICAL, ELECTRONIC`. 

## Steps to Reproduce
Go to the form to create a new borrowing request.
Fill in all the required fields.
For `item type` choose `Physical copy`.
Press the button to create borrowing request.

## The expected behavior
It should accept both of these fields.

## Screenshot
Selection_015

KonstantinaStoikou:
Closes #60


src/lib/pages/backoffice/ILL/BorrowingRequest/BorrowingRequestEditor/BorrowingRequestForm/BorrowingRequestForm.js
```diff
@@ -15,7 +15,7 @@ import { Payment } from './Payment';
 
 const submitSerializer = values => {
   const submitValues = { ...values };
-  submitValues.library_pid = values.library.pid;
+  submitValues.library_pid = values.library.id;
```
ntarocco, line 18: Maybe better:
```suggestion
  submitValues.library_pid = values.library.metadata.pid;
```
but you might have to check if it is set? (when creating a new one)

KonstantinaStoikou: I am not sure but I think this `id` is coming from here https://github.com/inveniosoftware/react-invenio-app-ils/blob/70ba2afcbcb433f3cb416070c06c50bd6c170483/src/lib/modules/ESSelector/serializer.js#L108 so I don't think I can change it.

Issue #73, opened 2020-06-18
ILL: library field empty on edit


kprzerwa:
where:
http://localhost:3000/backoffice/ill/borrowing-requests/<pid> (any pending borrowing request)

reproduce:
1. Edit the borrowing request
2. Change state to requested
3. Save

Observed incorrect behaviour
It shows form validation error on the library field even though it is not missing.

PR #75, opened 2020-06-19
UI bugfixing

kprzerwa:
to review: last 2 commits
closes: #73

### DIFF
--- a/src/lib/components/CancelModal/CancelModal.js
+++ b/src/lib/components/CancelModal/CancelModal.js
@@ -65,6 +65,7 @@ export default class CancelModal extends Component {
             content={buttonText}
             onClick={this.show}
             loading={isLoading}
+            disabled={isLoading}
           />
         }
         open={open}
--- a/src/lib/components/CancelModal/CancelModal.js
+++ b/src/lib/components/CancelModal/CancelModal.js
@@ -53,13 +53,20 @@ export default class CancelModal extends Component {
   };
 
   render() {
-    const { buttonText, cancelText, content, header } = this.props;
+    const { buttonText, cancelText, content, header, isLoading } = this.props;
     const { open, value, showPopup } = this.state;
     return (
       <Modal
         basic
         size="small"
-        trigger={<Button primary content={buttonText} onClick={this.show} />}
+        trigger={
+          <Button
+            primary
+            content={buttonText}
+            onClick={this.show}
+            loading={isLoading}
+          />
+        }
         open={open}
         onClose={this.hide}
       >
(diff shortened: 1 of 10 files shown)


src/lib/api/date.js
```diff
@@ -51,5 +51,5 @@ export const toShortDate = date => {
  */
 export function dateFormatter({ col, row }, defaultValue = null) {
   const dateField = _get(row, col.field);
-  return dateField ? toShortDate(fromISO(dateField)) : defaultValue;
+  return dateField ? fromISO(dateField).toLocaleString() : defaultValue;
```
ntarocco: 👍

src/lib/components/CancelModal/CancelModal.js
```diff
@@ -53,13 +53,20 @@ export default class CancelModal extends Component {
...
       <Modal
         basic
         size="small"
-        trigger={<Button primary content={buttonText} onClick={this.show} />}
+        trigger={
+          <Button
+            primary
+            content={buttonText}
+            onClick={this.show}
+            loading={isLoading}
```
ntarocco, line 67: Could you please:
```suggestion
            loading={isLoading}
            disabled={isLoading}
```
I think we should add to all buttons with a loading state: my fear (did not test it!) is that you can still click while loading? We should maybe test it...

src/lib/modules/Loan/actions.js
```diff
@@ -60,12 +60,7 @@ export const performLoanAction = (
         type: ACTION_SUCCESS,
         payload: response.data,
       });
-
-      dispatch({
-        type: DETAILS_SUCCESS,
-        payload: response.data,
-      });
-
+      dispatch(fetchLoanDetails(response.data.pid));
```
ntarocco, line 63:
```suggestion
      dispatch(fetchLoanDetails(response.data.medata.pid));
```
?

I would really love to remove that extra `pid` at the root level

src/lib/pages/backoffice/Loan/LoanDetails/LoanActions/LoanActions.js
```diff
@@ -32,9 +33,10 @@ export default class LoanActions extends Component {
               cancelText="Cancel Loan"
               buttonText="cancel"
               action={cancelAction}
+              isLoading={isLoading}
```
ntarocco, line 36: same here for the buttons disabled

src/lib/components/CancelModal/CancelModal.js
kprzerwa: yes, true, good point

PR #88, opened 2020-06-23
ui: make some components overridable for cds-books

KonstantinaStoikou:
- Make `SectionsWrapper` and`ILSMenu` overridable
- Display max 5 books in each section in homepage
- Export in `index.js` whatever is needed for cds-books ([PR for cds-books](https://github.com/CERNDocumentServer/cds-books/pull/104))


src/lib/pages/frontsite/Home/Sections/SectionsWrapper.js
```diff
@@ -70,11 +71,14 @@ export default class SectionsWrapper extends Component {
     const { sections } = this.props;
     if (!_isEmpty(sections)) {
       return (
-        <Container fluid className="fs-landing-page-section-wrapper">
-          {sections.map((Section, idx) => {
-            return <Section key={idx} className="fs-landing-page-section" />;
-          })}
-        </Container>
+        <Overridable id="SectionsWrapper.sections" sections={sections}>
```
kprzerwa: this should be change due to introducing react-invenio-app-ils we no longer need to keep the renderDefaultSection functions

src/lib/modules/Document/DocumentCardGroup/DocumentCardGroup.js
```diff
@@ -44,7 +44,7 @@ class DocumentCardGroup extends Component {
   renderCards() {
     const { data } = this.state;
     if (!data) return null;
-    return data.hits.map(book => {
+    return data.hits.slice(0, 5).map(book => {
```
kprzerwa, line 47: this should be passed as a prop, not hardcoded (the slice size)

src/lib/modules/Document/DocumentCardGroup/DocumentCardGroup.js
```diff
@@ -44,7 +44,7 @@ class DocumentCardGroup extends Component {
   renderCards() {
     const { data } = this.state;
     if (!data) return null;
-    return data.hits.map(book => {
+    return data.hits.slice(0, this.props.displayNumber).map(book => {
```
kprzerwa, line 47: syntax: the props should be deconstructed, you probably had a warning in VS code for this.

in general it would better if instead of slicing the data hits, you would pass the `displayNumber` prop as a parameter to `fetchDataMethod` which is an API call. By default it return 10, but might be set to more, so we use only a half of the objects in the response we get from the backend. If you pass the page size to the backend, you will  get back exactly the number you need - the response will be smaller and there will be no need to manipulate the output
you can check `withSize` in `borrowingRequest.js` for reference

src/lib/index.js
```diff
@@ -1,2 +1,6 @@
+export { documentApi } from '@api/documents';
```
ntarocco, line 1: It would be great if you export all APIs... basically:
* all APIs
* history
* App
* store
* the components you need for now

Can you please delete the file `src/lib/api/index.js` and fix any import if it breaks?

src/lib/pages/frontsite/Home/Sections/SectionsWrapper.js
```diff
@@ -25,6 +23,7 @@ export default class SectionsWrapper extends Component {
                 .query()
                 .withDocumentType('BOOK')
                 .sortBy('mostrecent')
+                .withSize(5)
```
kprzerwa, line 26: just a small thingy: the `5` should be coming from props, not to be hardcoded, like previously. You could add a prop which has a default equal 5

src/lib/pages/frontsite/Home/Sections/SectionInstallation.js
```diff
@@ -126,6 +126,19 @@ export class SectionInstallation extends Component {
                               </List.Item>
                             </List>
                           </List.Item>
+                          <List.Item className="install-item">
+                            Test ILS as a librarian by logging in with
+                            <List>
+                              <List.Item>
+                                <List.Header>
+                                  email: librarian@test.ch
```
ntarocco, line 134: 👍

PR #93, opened 2020-07-01
fix static pages config

KonstantinaStoikou:


src/index.js
```diff
@@ -6,39 +7,18 @@ import { OverridableContext } from 'react-overridable';
...
 
 const config = {
   invenioConfig: {
     authors: {
       maxDisplay: 42,
     },
   },
-  uiConfig: {},
-  searchConfig: {},
+  uiConfig: { ...uiConfig },
```
kprzerwa, line 20: this will be probably refactored, concerns the problems we talked about on the standup today

src/lib/routes/frontsite/Frontsite.js
```diff
@@ -23,7 +23,8 @@ export default class FrontSite extends Component {
   };
 
   render() {
-    const staticPagesRoutes = getStaticPagesRoutes();
+    const { config } = this.props;
+    const staticPagesRoutes = getStaticPagesRoutes(config.uiConfig.staticPages);
```
kprzerwa, line 27: not a blocker: the config can be deconstructed (probably it show ESLint warning for this line):

```
const {config: {uiConfig: {staticPages}} };
```

src/index.js
```diff
@@ -6,39 +7,18 @@ import { OverridableContext } from 'react-overridable';
 import { Router } from 'react-router-dom';
 import 'semantic-ui-less/semantic.less';
 import { InvenioILSApp } from './lib';
+import { searchConfig } from './lib/config/searchConfig';
 
-// const CustomHome = ({ ...props }) => {
```
topless, line 10: nit: @ntarocco suggested in a previous review, and I would agree, it was a good idea to have them there as an example.

src/index.js
topless: What about `uiConfig: uiConfig`?

src/lib/App.js
```diff
@@ -75,9 +75,5 @@ export default class App extends Component {
 }
 
 App.propTypes = {
-  config: PropTypes.object,
-};
-
-App.defaultProps = {
-  config: {},
+  config: PropTypes.object.isRequired,
```
topless, line 78: As I understand this, your Application will not start if you don't provide a config.
IMHO it should be required if we have any required config settings for the App to work, and these props should be explicitly specified in expected PropTypes. If not then it be just empty `{}` if the used doesn't provide one.

src/lib/components/ILSFooter/ILSFooter.js
```diff
@@ -16,10 +16,21 @@ const FooterLayout = props => {
                 <Header as="h4" content="More information" />
                 <List>
                   <List.Item>
-                    <Link to={getStaticPageByName('about').route}>About</Link>
+                    <Link
+                      to={
+                        getStaticPageByName(uiConfig.staticPages, 'about').route
```
topless, line 21: I see we are passing `uiConfig.staticPages` every time we call the explicit functions `getStaticPageByName`.
What is the case for that? I think it should be `getStaticPagesByName('about')`?

*same comment for the rest of `getStaticPage*`

src/lib/components/ILSFooter/ILSFooter.js
```diff
@@ -1,4 +1,4 @@
-import { getStaticPageByName } from '@config/uiConfig';
+import { getStaticPageByName, uiConfig } from '@config/uiConfig';
```
topless, line 1: nit: maybe the `uiConfig` should be imported from `@config` as everywhere else.

src/lib/routes/frontsite/Frontsite.js
```diff
@@ -77,6 +78,7 @@ export default class FrontSite extends Component {
 
 FrontSite.propTypes = {
   customStaticPages: PropTypes.func,
+  config: PropTypes.object.isRequired,
```
topless, line 81: Same as before why required? Sorry if I am busting nerves..

src/index.js
kprzerwa: does it make a difference now since it will be refactored anyway?

src/lib/components/ILSFooter/ILSFooter.js
kprzerwa: the approach you are proposing didn't work because of the mutated config problem, it was getting the ILS static pages here instead of CDS ones

src/lib/components/ILSFooter/ILSFooter.js
kprzerwa: if we abuse too much of these collective index exports it might lead to circular dependencies and if we have a habit of importing from exact location it is less likely to happen

src/lib/routes/frontsite/Frontsite.js
kprzerwa: IMO without the config the app will not work, that is why it should be mandatory

src/index.js
kprzerwa: we have it overwritten now in cds-books anyway, and react-overridable has examples in the documentation so IMO we can clean it up now

src/lib/components/ILSFooter/ILSFooter.js
topless: I will handle it in my upcoming pr 👍

src/lib/routes/frontsite/Frontsite.js
topless: Perfect! What is missing? For example if a `title` is required this config should be a shape with a title required.

topless: Let's go philosophical on this one, what should be required for an App to work?

PR #95, opened 2020-07-07
document: change keywords field type to array

kprzerwa:


ntarocco:
See my comment in the backend PR

PR #100, opened 2020-07-10
docreq: add medium field

### LINKED ISSUE #99: Document request: add `medium` fields
When the patron request a new book that is not in the library catalogue, (s)he will have to fill the Document Request form. The form needs to be enhanced with the new field medium and values Paper or Print and E-book or Electronic.

## Tasks

- [  ] change the existing form and add the new dropdown (select) with the possible values
- [  ] make sure the new field is displayed in the frontiste profile page and in the backoffice
- [  ] improve the text `Please fill in the request form to request a new literature from the library.` to explain that login will be requested

Backend: https://github.com/inveniosoftware/invenio-app-ils/issues/633

jrcastro2:
* adds medium field to frontsite
  profile page and backoffice

* adds new dropdown to the
  request document form 

* closes #99 

image
--------------------------------
image
--------------------------------
image

### DIFF
--- a/src/lib/forms/core/SelectField.js
+++ b/src/lib/forms/core/SelectField.js
@@ -1,8 +1,8 @@
-import React, { Component } from 'react';
-import PropTypes from 'prop-types';
 import { FastField, Field, getIn } from 'formik';
-import { Form } from 'semantic-ui-react';
 import _isEmpty from 'lodash/isEmpty';
+import PropTypes from 'prop-types';
+import React, { Component } from 'react';
+import { Form } from 'semantic-ui-react';
 
 export class SelectField extends Component {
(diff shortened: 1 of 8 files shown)


src/lib/forms/core/SelectField.js
```diff
@@ -82,7 +81,7 @@ export class SelectField extends Component {
       options,
       ...uiProps
     } = this.props;
-    const value = getIn(values, fieldPath, multiple ? [] : defaultValue);
+    const value = getIn(values, fieldPath, multiple);
```
ntarocco, line 84: My mistake in suggesting you this change, it looks like value is used here, we might want to simply do:
```
     const value = getIn(values, fieldPath, multiple ? [] : ''); 
```
The other changes look ok to me

PR #101, opened 2020-07-13
docreq: Adds request type field

jrcastro2:
* Adds request type field to frontsite
  profile page and backoffice

* Adds 2 new dropdown and a string field
  to the request document form

* Adds payment method and payment
  information to backoffice

image
---------
image

---------
image

### DIFF
--- a/src/lib/config/invenioConfig.js
+++ b/src/lib/config/invenioConfig.js
@@ -211,7 +211,9 @@ export const invenioConfig = {
       ill_item_type: 'ill_item_type',
       ill_payment_mode: 'ill_payment_mode',
     },
-    docReq: {
+    documentRequests: {
+      doc_req_type: 'doc_req_type',
+      doc_req_payment_method: 'doc_req_payment_method',
       doc_req_medium: 'doc_req_medium',
     },
   },
--- a/src/lib/forms/core/StringField.js
+++ b/src/lib/forms/core/StringField.js
@@ -53,5 +53,5 @@ StringField.propTypes = {
 StringField.defaultProps = {
   inline: false,
   optimized: true,
-  width: null,
+  width: 16,
 };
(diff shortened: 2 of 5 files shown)


src/lib/forms/core/StringField.js
```diff
@@ -47,9 +47,11 @@ StringField.propTypes = {
   fieldPath: PropTypes.string.isRequired,
   inline: PropTypes.bool,
   optimized: PropTypes.bool,
+  width: PropTypes.number,
 };
 
 StringField.defaultProps = {
   inline: false,
   optimized: true,
+  width: null,
```
topless, line 56: `null` will shrink the String field to its default size which means if there are a few components in one row, they will not expand to cover it all.

Screenshot 2020-07-14 13 40 06

We could use the maximum (16) number instead so it will `push` the rest of the elements to fill the row, and if its on its own it will fill the row.

Screenshot 2020-07-14 13 40 17

src/lib/config/invenioConfig.js
```diff
@@ -211,6 +211,10 @@ export const invenioConfig = {
       ill_item_type: 'ill_item_type',
       ill_payment_mode: 'ill_payment_mode',
     },
+    docReq: {
```
topless, line 214: For consistency use `documentRequests` instead of `docReq`, in general we try to avoid different abbreviations.

src/lib/pages/frontsite/PatronProfile/PatronCurrentDocumentRequests/PatronCurrentDocumentRequests.js
```diff
@@ -207,6 +207,12 @@ class PatronCurrentDocumentRequests extends Component {
     { title: 'Created', field: 'created', formatter: dateFormatter },
     { title: 'Authors', field: 'metadata.authors' },
     { title: 'Publication year', field: 'metadata.publication_year' },
+    {
+      title: 'Request type',
```
topless, line 211: I noticed in the screenshot that the request type appears in capital `BUY`, in other cases we introduce labels so we have full control on what to display per field.

topless:
Very nicely done 🎉 !! A couple of small observations, ping me if needed!

src/lib/pages/frontsite/PatronProfile/PatronCurrentDocumentRequests/PatronCurrentDocumentRequests.js
kprzerwa: I agree. Also `Purchase` sounds better than `Buy`

src/lib/forms/core/StringField.js
kprzerwa: yes you need to remember that this component is reused, and whatever you provide as a default will have the impact on all the other usages, which shouldn't be broken after your change

PR #104, opened 2020-07-14
bugs: relation components

### LINKED ISSUE #106: bug: create Edition relation
This is probably for [invenio-app-ils](https://github.com/inveniosoftware/invenio-app-ils), when in document and try to create an `edition` relation with any other type of document or series, I get an error 

```Cannot create a relation `edition` between PID `jnqws-cdf46` and PID `1zvmt-vr708`, they are different record types```

### PR DESCRIPTION
**Pull request by topless**
- fixes RelationSerial component
- fixes RelationEdition component
- fixes propTypes for anchors
- fixes Sidebar html hierarchy
- fixes components with required props to render only if are available
- fixes keys for AlternativeTitles
- fixes connections with reducers
- fixes render conditions, check for empty arrays instead of negating the length ![].length 
- fixes LiteratureRelations for series
- passed recordDetails in common relation components
- closes #106

### CODE REVIEW

File: src/lib/pages/frontsite/Series/SeriesDetails/SeriesPanel/SeriesPanel.js [hunk 1]
```diff
@@ -34,10 +35,9 @@ class SeriesPanel extends Component {
                 </Grid.Column>
                 <Grid.Column width={6}>
                   <ILSHeaderPlaceholder isLoading={isLoading}>
-                    <SeriesTitle
-                      title={series.metadata.title}
-                      modeOfIssuance={series.metadata.mode_of_issuance}
-                    />
+                    {title && moi && (
```
kprzerwa, line 38: as the moi is required, the title too, you could just do 
```
isEmpty(series.metadata) && (
```

src/lib/pages/frontsite/Series/SeriesDetails/SeriesPanel/SeriesPanel.js
```diff
@@ -61,7 +61,9 @@ class SeriesPanel extends Component {
                 </Grid.Column>
                 <Grid.Column width={5}>
                   <SeriesAccess urls={series.metadata.access_urls} />
-                  <SeriesSequences relations={series.metadata.relations} />
+                  {series.metadata.relations && (
```
kprzerwa, line 64: isn't it the `series.metadata.relations` an Object? I think in this case it will be always evaluated as true

src/lib/pages/frontsite/Series/SeriesDetails/SeriesPanel/SeriesPanel.js
topless: yes but this check will not fail if you have one of the two, since it will not be empty. To be honest I had it `!isLoading` as a condition for its render, but switched it to explicit `title` and `moi` so you don't have to dive further to check what is required since its already there. No really strong opinion.

src/lib/modules/Literature/LiteratureRelations.js
```diff
@@ -22,9 +22,7 @@ export default class LiteratureRelations extends Component {
 
   renderMultiparts = () => {
     const relations = _get(this.relations, 'multipart_monograph', []);
-    if (!relations.length) {
-      return null;
-    }
+    if (_isEmpty(relations)) return null;
```
ntarocco, line 25: not sure I understand this change: if undefined, it will be set to `[]`. Did you found issues with all this?

src/lib/pages/backoffice/Document/DocumentDetails/DocumentContent.js
```diff
@@ -2,8 +2,10 @@ import { DocumentEItems } from '@pages/backoffice/Document/DocumentDetails/Docum
 import { DocumentItems } from '@pages/backoffice/Document/DocumentDetails/DocumentItems';
 import { DocumentPendingLoans } from '@pages/backoffice/Document/DocumentDetails/DocumentPendingLoans';
 import { DocumentStats } from '@pages/backoffice/Document/DocumentDetails/DocumentStats';
-import DocumentSeries from '@pages/backoffice/Document/DocumentDetails/DocumentRelations/DocumentSeries';
-import DocumentSiblings from '@pages/backoffice/Document/DocumentDetails/DocumentRelations/DocumentSiblings';
+import {
+  DocumentSeries,
+  DocumentSiblings,
+} from '@pages/backoffice/Document/DocumentDetails/DocumentRelations';
```
ntarocco, line 8: just for info, we have tried to remove the `index.js`, whenever possible...

src/lib/pages/frontsite/Series/SeriesDetails/SeriesPanel/SeriesPanel.js
ntarocco: not sure I understand, they are both required no? What is the issue here?

src/lib/pages/frontsite/Series/SeriesDetails/SeriesPanel/SeriesSequences.js
```diff
@@ -27,20 +28,18 @@ class SeriesSequences extends Component {
 
   render() {
     const { relations } = this.props;
-
     const sequenceRelations = _get(relations, 'sequence', []);
-    const hasSequenceRelations = sequenceRelations.length > 0;
+    if (_isEmpty(sequenceRelations)) return null;
```
ntarocco, line 32: I still prefer to use native JS instead of abusing of `isEmpty`, do you really need it here?

src/lib/components/backoffice/Sidebar/Sidebar.js
```diff
@@ -202,38 +202,35 @@ class Sidebar extends Component {
...
+                  <Label.Detail>
+                    {process.env.REACT_APP_INVENIO_VERSION}
+                  </Label.Detail>
+                </Label>
+              </List.Item>
+              {process.env.REACT_APP_OVERLAY_VERSION && (
+                <List.Item>
                   <Label color="blue">
-                    invenio-ils
+                    overlay
```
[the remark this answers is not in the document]
topless: https://github.com/inveniosoftware/react-invenio-app-ils/blob/4aede755893278150d0ec856716841259e2a41b5/src/lib/components/backoffice/Sidebar/Sidebar.js#L208

The problem here, is that `<Label>` which introduces a div, is wrapped in a `<p>` element, and it is invalid html.

src/lib/pages/backoffice/Document/DocumentDetails/DocumentContent.js
topless: These two are the only case being exported from the same `index.js`, for the rest of the *Relation* components with their modals they are exported from their index in order to connect them with the reducer. We can even deprecate completely the `index.js` and connect them in their source file, which would also prevent us from importing from the wrong place, i.e this case where the components were imported explicitly though, not connected to their reducers.

src/lib/modules/Literature/LiteratureRelations.js
topless: you are right on these ones I change them pro-actively to avoid comparison with `undefined` as you mentioned. I revert

src/lib/pages/frontsite/Series/SeriesDetails/SeriesPanel/SeriesPanel.js
topless: If we don't prevent the component from rendering when they don't exist we get errors, I believe this is the way to use the `required` propTypes.

src/lib/pages/frontsite/Series/SeriesDetails/SeriesPanel/SeriesSequences.js
topless: this one slipped, thanks, fixed.

src/lib/pages/frontsite/Series/SeriesDetails/SeriesPanel/SeriesPanel.js
ntarocco: yes perfect, agree, but `title` and `moi` are mandatory in the JSON schema, how come they are not defined?

topless: @ntarocco our [`initialState`](https://github.com/inveniosoftware/react-invenio-app-ils/blob/4aede755893278150d0ec856716841259e2a41b5/src/lib/pages/frontsite/Series/SeriesDetails/state/reducer.js#L6) in the reducer is `data: { metadata: {} }`, and goes `SeriesDetail` -> `SeriesPanel` -> `SeriesTitle`.

We had something like the following 

`<SeriesTitle title={series.metadata.title} modeOfIssuance={series.metadata.mode_of_issuance} />`

which would spit warnings like

```
Warning: Failed prop type: The prop `title` is marked as required in `SeriesTitle`, but its value is `undefined`.
```

kprzerwa: yes, is this short moment when we do not have yet the data and the components use the initial state from the reducer
one of the solutions could be to provide the initial object with the required fields, but it seems problematic to maintain (in case of changes we need to remember to go and change the initial state too), however it seems safer to fail on the initial state

PR #112, opened 2020-07-15
Backoffice: make components overridable for delivery method icon

KonstantinaStoikou:
Overridable components for this: https://github.com/CERNDocumentServer/cds-books/pull/117


src/lib/pages/backoffice/Document/DocumentDetails/state/actions.js
```diff
@@ -144,3 +146,31 @@ export const requestLoanForPatron = (
...
+    );
+  }
+}
+
+RequestForPatronMessage.propTypes = {
+  deliveryMethod: PropTypes.string.isRequired,
+  loanPid: PropTypes.string.isRequired,
+};
+
+export default Overridable.component(
```
ntarocco, line 173: Why not adding a prop, `icon`, instead of `deliveryMethod` and use the `parametrize` overriding? Probably hard to explain here, writing, let's chat tomorrow!

PR #121, opened 2020-07-21
fix displaying keywords

kprzerwa:


src/lib/modules/Document/DocumentInfo.js
```diff
@@ -36,6 +36,7 @@ export class DocumentInfo extends Component {
               ? keywordValue
               : '';
           })}
+          {keywords.length < 1 && 'No keywords provided.'}
```
ntarocco, line 39: Just for info, not to be changed, you can also do:
```suggestion
          {!keywords.length && 'No keywords provided.'}
```

kprzerwa: nice, I will use this from now on

PR #122, opened 2020-07-21
backoffice: add loading state to loan checkout

### LINKED ISSUE #77: backoffice - loan details - loading state missing
Screenshot 2020-06-19 at 10 22 52
Where: pending loan
Description: After clicking on "Checkout item" the button should have loading and disabled  state so the user can't click it twice by mistake

After the loan is correctly created the checkout buttons should be disabled

FlorianCassayre:
* closes #77

image

### DIFF
--- a/src/lib/pages/backoffice/Loan/LoanDetails/AvailableItems/AvailableItems.js
+++ b/src/lib/pages/backoffice/Loan/LoanDetails/AvailableItems/AvailableItems.js
@@ -63,11 +63,13 @@ export default class AvailableItems extends Component {
   }
 
   checkoutItemButton(item, loan) {
-    const { performCheckoutAction } = this.props;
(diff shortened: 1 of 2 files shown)


src/lib/pages/backoffice/Loan/LoanDetails/AvailableItems/AvailableItems.js
```diff
@@ -64,11 +65,15 @@ export default class AvailableItems extends Component {
 
   checkoutItemButton(item, loan) {
     const { performCheckoutAction } = this.props;
+    const { isLoading } = this.state;
```
kprzerwa, line 68: this should be mapped via redux with ACTION_IS_LOADING of /modules/Loan/actions.js, not with the internal state of the component, otherwise if the checkout fails, the page will not reload and the button will be stuck in loading state

PR #123, opened 2020-07-22
backoffice: add identifiers to e-items

FlorianCassayre:
* new field in create/update form
* identifiers appear in the metadata detail panel
* requires inveniosoftware/invenio-app-ils#914
* closes inveniosoftware/invenio-app-ils#658

image

image

### DIFF
--- a/src/lib/pages/backoffice/EItem/EItemDetails/EItemMetadata/EItemMetadata.js
+++ b/src/lib/pages/backoffice/EItem/EItemDetails/EItemMetadata/EItemMetadata.js
@@ -4,6 +4,7 @@ import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 import ShowMore from 'react-show-more';
 import { Grid, Header, Segment } from 'semantic-ui-react';
+import { IdentifierRows } from '@modules/Identifiers';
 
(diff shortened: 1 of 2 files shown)


FlorianCassayre:
I noticed there was a bug with the scheme dropdown: the current value (as provided by the server) is not selected accordingly on the UI. Is there an open issue about this problem?

src/lib/pages/backoffice/EItem/EItemDetails/EItemMetadata/EItemMetadata.js
```diff
@@ -19,6 +20,12 @@ export default class EItemMetadata extends Component {
         name: 'Available urls',
         value: <UrlList urls={eitemDetails.metadata.urls} />,
       },
+      {
+        name: 'Identifiers',
+        value: (
+          <IdentifierRows identifiers={eitemDetails.metadata.identifiers} />
```
kprzerwa, line 26: did you try with empty identifiers field ? or with the field not existing on the object?

FlorianCassayre: (fixed in `26c037c`)

PR #131, opened 2020-07-27
global: minor fixes

### LINKED ISSUE #42: When pid not found it should display 404 page
Software version (if known):

## Describe the bug

The 404 error from the bakcend is not properly handled

## Steps to Reproduce

1. Visit /documents/43255
2. See the Error notification

## The current behavior
It displays a notification for an error.

## The expected behavior
It should display 404 page.

### LINKED ISSUE #108: remove Restricted facet from the patron search page
The facet `Restricted` should never appear in the patrons search.

This might be a result of a  bigger problem of wrongly assigned model types in `SearchControls` components. Review them **carefully**.

If not a problem with configuration, implement a solution to hide the filter when the user has no librarian or admin role or (good enough if the first case not possible) or leave the filter only for the backoffice search.

### LINKED ISSUE #109: rename relation filter values - document search
Screenshot 2020-07-22 at 15 51 02
Rename the filter title and/or values for something which is more user friendly. 
From the current state the user does not know that the filter is supposed to filter the documents which have other related editions/languages or are a part of series.

FlorianCassayre:
* displays 404 page when literature or series is not found on the frontsite (closes #42)
* removed "Restricted" facet from the frontsite search (closes #108)
* support for configurable search filters labels, renamed the "Relation" filter (closes #109)
* repaired link to location editor

### DIFF
--- a/src/lib/config/common.js
+++ b/src/lib/config/common.js
@@ -27,3 +27,12 @@ export const ITEM_MEDIUMS = [
   { value: 'PAPERBACK', text: 'Paperback' },
   { value: 'HARDCOVER', text: 'Hardcover' },
 ];
+
+export const DOCUMENT_RELATIONS = [
+  { value: 'edition', text: 'By edition' },
+  { value: 'multipart_monograph', text: 'By series (completed)' },
+  { value: 'serial', text: 'By series (periodic)' },
+  { value: 'language', text: 'By language' },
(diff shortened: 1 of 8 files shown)


FlorianCassayre:
About the `404` error patch, I realized that `hasError === true` does not always imply that `error.response` is defined (for instance if server is down or if host cannot be resolved).

PR #133, opened 2020-07-27
auth: fix providers config for display

kprzerwa:


ntarocco:
I need another couple of fixes, can you wait to merge? I can push here...

PR #139, opened 2020-07-28
loanreq: checkout by scan

### LINKED ISSUE #44: Loan request: checkout by scan
## Use case
Librarian selects loan request (pending) to be fullfilled. Gets the book from the shelf and proceeds to check-out. Scans the book to fulfil the loan.

## Description
On the loan request detail page, besides the `Check-out` button next to each available item, we should have a search field with barcode scan functionality, to avoid clicking on the wrong check-out button.

jrcastro2:
* Adds the option to scan a copy
  to filter the available physical
  copies list
* closes #44
### Loan details page
image
-----------------------------
### Filtering results
image
-----------------------------
### No results found
image

### DIFF
--- a/src/lib/pages/backoffice/Loan/LoanDetails/AvailableItems/AvailableItems.js
+++ b/src/lib/pages/backoffice/Loan/LoanDetails/AvailableItems/AvailableItems.js
@@ -1,16 +1,18 @@
+import { itemApi } from '@api/items';
+import { recordToPidType } from '@api/utils';
+import { SeeAllButton } from '@components/backoffice/buttons/SeeAllButton';
 import { InfoMessage } from '@components/backoffice/InfoMessage';
+import { Error } from '@components/Error';
+import { Loader } from '@components/Loader';
 import { ResultsTable } from '@components/ResultsTable/ResultsTable';
+import { SearchBar } from '@components/SearchBar';
+import { invenioConfig } from '@config';
+import { BackOfficeRoutes } from '@routes/urls';
+import _isEmpty from 'lodash/isEmpty';
+import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 import { Link } from 'react-router-dom';
-import PropTypes from 'prop-types';
 import { Button, Header, Message, Segment } from 'semantic-ui-react';
-import { Loader } from '@components/Loader';
-import { Error } from '@components/Error';
-import { itemApi } from '@api/items';
-import { invenioConfig } from '@config';
-import { SeeAllButton } from '@components/backoffice/buttons/SeeAllButton';
-import { BackOfficeRoutes } from '@routes/urls';
-import { recordToPidType } from '@api/utils';
 
 export default class AvailableItems extends Component {
   constructor(props) {
@@ -19,6 +21,7 @@ export default class AvailableItems extends Component {
     this.assignItemToLoan = props.assignItemToLoan;
     this.performCheckoutAction = props.performCheckoutAction;
     this.seeAllUrl = BackOfficeRoutes.itemsListWithQuery;
+    this.state = { query: '', filteredData: null };
   }
 
   componentDidMount() {
@@ -49,6 +52,7 @@ export default class AvailableItems extends Component {
     return (
       <Button
         size="mini"
+        color="blue"
         onClick={() => {
           const itemPid = {
             type: recordToPidType(item),
@@ -57,7 +61,7 @@ export default class AvailableItems extends Component {
           assignItemToLoan(itemPid, pid);
         }}
       >
-        replace current physical copy
+        replace
       </Button>
     );
   }
@@ -67,7 +71,7 @@ export default class AvailableItems extends Component {
     return (
       <Button
         size="mini"
-        color="teal"
+        color="blue"
         loading={isActionLoading}
         disabled={isActionLoading}
         onClick={() => {
@@ -88,6 +92,7 @@ export default class AvailableItems extends Component {
 
   renderTable() {
     const { data } = this.props;
+    const { filteredData } = this.state;
     const columns = [
       { title: 'PID', field: 'metadata.pid' },
       {
@@ -112,7 +117,7 @@ export default class AvailableItems extends Component {
 
     return (
       <ResultsTable
-        data={data.hits}
+        data={filteredData === null ? data.hits : filteredData}
         columns={columns}
         totalHitsCount={data.total}
         name="available physical copies"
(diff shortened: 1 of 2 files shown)


src/lib/pages/backoffice/Actions/CheckIn/ItemsSearch/ItemsSearch.js
```diff
@@ -16,13 +17,35 @@ export default class ItemsSearch extends Component {
...
   };
 
+  performCheckout(queryString) {
+    const { checkoutData, loan, performCheckoutAction } = this.props;
+    checkoutData.hits.forEach(hit => {
+      if (hit.metadata.barcode === queryString) {
+        this.props.loadingHandler();
+        const checkoutUrl = loan.availableActions.checkout;
+        const patronPid = loan.metadata.patron_pid;
+        const documentPid = hit.metadata.document.pid;
```
ntarocco, line 37: Just to be on the safe side:
```suggestion
        const documentPid = hit.metadata.document_pid;
```
* `document_pid` is a real field of the loan
* `document` is a `$ref`, resolved by the backend
I would prefer the first whenever possible

src/lib/pages/backoffice/Actions/CheckIn/ItemsSearch/ItemsSearch.js
```diff
@@ -50,11 +73,15 @@ export default class ItemsSearch extends Component {
           currentQueryString={queryString}
           onInputChange={this.onInputChange}
           executeSearch={this.onKeyPressHandler}
-          placeholder="Scan physical copy barcode to check-in..."
+          placeholder={
+            isCheckout
+              ? 'Scan physical copy barcode to checkout...'
```
ntarocco, line 78:
```suggestion
              ? 'Insert physical copy barcode to checkout...'
```
and also below. The thing is that we don't know if librarian will really scan, it depends on the libraries equipment.
It was already there, we missed it... can you please make sure there is no "scan" mentioning around?

src/lib/pages/backoffice/Loan/LoanDetails/AvailableItems/AvailableItems.js
```diff
@@ -109,16 +125,25 @@ export default class AvailableItems extends Component {
...
+          data={data.hits}
+          columns={columns}
+          totalHitsCount={data.total}
+          name="available physical copies"
+          seeAllComponent={this.seeAllButton()}
+          renderEmptyResultsElement={this.renderEmptyResults}
+        />
+        <LoanScanCopy
+          isOpened={scanModalIsOpen}
+          headerContent="Please scan the copy"
```
ntarocco, line 141:
```suggestion
          headerContent="Checkout"
```

src/lib/pages/backoffice/Loan/LoanDetails/AvailableItems/AvailableItems.js
```diff
@@ -155,12 +180,33 @@ export default class AvailableItems extends Component {
...
+      <Button
+        floated="right"
+        compact
+        size="tiny"
+        color="orange"
+        loading={isActionLoading}
+        disabled={isActionLoading}
+        onClick={this.openModal}
+      >
+        Scan a copy
```
ntarocco, line 195:
```suggestion
        Checkout with barcode
```

src/lib/pages/backoffice/Loan/LoanDetails/AvailableItems/AvailableItems.js
```diff
@@ -155,12 +180,33 @@ export default class AvailableItems extends Component {
     }
   };
 
+  scanCopyButton() {
```
ntarocco, line 183:
```suggestion
  openModalCheckoutWithBarcode() {
```

jrcastro2:
> Thanks for this feature! Looks great! I like the "Just a second"! :)
> 
> I have a major comment, sorry!!! The librarian would like to minimize clicks... What about if instead of opening a modal, the search bar is always showed, with focus, and on scan will filter the list of results? We can discuss IRL.

Currently the librarian will click the "scan a copy" button and the modal opens, now he can scan the copy and it will automatically checkout (if the barcode belongs to an available physical copy). 
If we have the search bar and it filters the list of result we still will have to click the checkout button of the copy. It would be the same amount of clicks, or am I missing something?
One way to reduce clicks would be to have the search bar as you say and on scan automatically checkout the copy. WDYT?

src/lib/pages/backoffice/Loan/LoanDetails/LoanScanCopy/LoanScanCopy.js
```diff
@@ -0,0 +1,104 @@
...
+            loadingHandler={this.isLoadingHandler}
+            errorHandler={this.wrongBarcodeHandler}
+          />
+        </Modal.Content>
+        <Modal.Description>
+          {this.state.isLoading
+            ? this.scanLoading()
+            : this.state.wrongBarcode
+            ? this.scanError()
+            : null}
```
FlorianCassayre, line 85: I would move this into `<Modal.Content>`, it looks weird without the margin.

(proposal)
image

jrcastro2: Thanks for noticing (for some reason I did not find it weird until now). Unfortunately the PR has changed and this will not be applied. Thanks anyway!

src/lib/pages/backoffice/Loan/LoanDetails/AvailableItems/AvailableItems.js
```diff
@@ -155,13 +167,52 @@ export default class AvailableItems extends Component {
...
+        executeSearch={this.onSearchExecute}
+        placeholder="Insert barcode..."
+      />
+    );
+  }
+
   render() {
-    const { isLoading, error } = this.props;
+    const { isLoading, error, loan, data } = this.props;
+    const statesThatRenderSearchBar = invenioConfig.CIRCULATION.loanActiveStates.concat(
```
ntarocco, line 204: I would display the box `Available physical copies` only for loans in `request` status? It is already with an active loan, then it should already have an item. WDYT?

jrcastro2: Currently the box offers the possibility to replace the current physical copy with another one available in case the item is on loan, that's why the search bar displays there too, to be able to filter in case the librarian wants to.
image
If the loan is cancelled or the item is returned the box still is displayed, I think it does not give any relevant information there. IMHO we could remove it in that case. WDYT?
image

PR #142, opened 2020-07-29
global: minor tweaks + patron page statistics loading

### LINKED ISSUE #94: remove the comma from list components when only one list entry
Example, in the list of languages, when only one entry, the comma should not be there:
Screenshot 2020-07-03 at 10 33 00

### LINKED ISSUE #105: Button should be disabled and in loading state after performing checkout action
Where: Backoffice item details page
Screenshot 2020-07-14 at 15 04 52

What should happen:
Before choosing the patron the checkout button should be disabled.
After choosing the patron the checkout button should be enabled.
After clicking "Checkout physical copy" all the visible checkout buttons should be in disabled and loading state until the action finishes with the response. If there is no error, the checkout button should be blocked with the information that the item is already on loan an it cannot be checked out.

### LINKED ISSUE #129: config: item medium does not match server specification
Mismatch for the `ONLINE` value:

https://github.com/inveniosoftware/react-invenio-app-ils/blob/master/src/lib/config/common.js#L22

https://github.com/inveniosoftware/invenio-app-ils/blob/master/invenio_app_ils/items/api.py#L116

Which one is right?

FlorianCassayre:
* fixed a typo
* added a prop to selector modal to disallow empty selections (closes #105)
* fixed config client/server mismatch (closes #129)
* fixed trailing comma in enumeration (closes #94)
* patron page: added loading state to statistics

(loading state)
image

(final state; note that the alignment stays the same)
image

### DIFF
--- a/src/lib/pages/frontsite/PatronProfile/PatronOverview/PatronOverview.js
+++ b/src/lib/pages/frontsite/PatronProfile/PatronOverview/PatronOverview.js
@@ -1,39 +1,42 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
-import { Container, Statistic } from 'semantic-ui-react';
+import { Container, Loader, Statistic } from 'semantic-ui-react';
 
 export default class PatronOverview extends Component {
+  renderStatistic(stat, quantifiableLabel) {
+    const {
+      isLoading,
+      data: { total },
+    } = stat;
+    return (
+      <Statistic>
+        <Statistic.Value>
+          {isLoading ? <Loader active inline /> : total}
+        </Statistic.Value>
+        <Statistic.Label>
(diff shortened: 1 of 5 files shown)


src/lib/pages/frontsite/PatronProfile/PatronOverview/PatronOverview.js
```diff
@@ -1,39 +1,42 @@
...
+      isLoading,
+      data: { total },
+    } = stat;
+    return (
+      <Statistic>
+        <Statistic.Value>
+          {isLoading ? <Loader active inline /> : total}
+        </Statistic.Value>
+        <Statistic.Label>
+          {isLoading ? <>&nbsp;</> : quantifiableLabel(total !== 1)}
```
ntarocco, line 17: 👍

PR #143, opened 2020-07-29
Datepicker: fix mobile view

### LINKED ISSUE #127: Datepicker component is broken on mobile devices
The days columns are wrongly displayed on top of each other on mobile devices only.

image

KonstantinaStoikou:
closes #127 
Screenshot from 2020-07-29 14-29-00

### DIFF
--- a/src/semantic-ui/site/collections/table.overrides
+++ b/src/semantic-ui/site/collections/table.overrides
@@ -2,8 +2,9 @@
 Table Overrides - REACT-INVENIO-APP-ILS
 *******************************/
 
-.bo-relations-tab, .bo-metadata-tab, .bo-metadata-segment {
-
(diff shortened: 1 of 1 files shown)


src/semantic-ui/site/collections/table.overrides
```diff
@@ -65,27 +65,41 @@ Table Overrides - REACT-INVENIO-APP-ILS
   }
 }
 
-
 @media only screen and (max-width: @largestMobileScreen) {
+  .ui.popup.transition {
+    table {
```
kprzerwa, line 70: thanks for the fix! it will be easier to fix this by applying (:not) selector to the table below, as in 
```
table:not(.ui.popup.transition) { ...
```
it will reduce the amount of css

PR #148, opened 2020-08-03
search: upgrade react-searchkit components

### LINKED ISSUE #4: upgrade search components to newest version of react-searchkit
The search components still have `renderElement=` props, and are not overridable. Create overrides of these components on the level of this ui library.
 Create the documentation as they will slightly different than other overridables of the invenio-app-ils - as they are overriden on react-searchkit level.

In this case it means that the default components will be used, and for recreating the look and feel of the current ones they need to be injected on the App (Context provider) level

they will be ovewritten in the overlay too

kprzerwa:
* closes #4


src/lib/modules/Literature/LiteratureSearchResultsGrid.js
```diff
@@ -2,39 +2,46 @@ import { recordToPidType } from '@api/utils';
...
-            {cards}
-          </Card.Group>
-        </Responsive>
-        <Responsive maxWidth={Responsive.onlyLargeScreen.minWidth - 1}>
-          <Card.Group doubling stackable itemsPerRow={3}>
-            {cards}
-          </Card.Group>
-        </Responsive>
-      </>
+export const LiteratureSearchResultsGrid = ({
```
zzacharo, line 9: Do you have now a rule on when to create functional components?

kprzerwa: not really :) usually I do when the component is stateless and simple

PR #150, opened 2020-08-05
mobile facets and sort order fix

### LINKED ISSUE #16: frontsite: Fix facets in the mobile view
after 0.15.0 version of react-searchkit, fix the facets display of document search
- [ ] the filters should be correctly rendered , f.e. as dropdowns
- [ ] if it is a boolean -> toggle filter

kprzerwa:
* closes #16 
Screenshot 2020-08-05 at 11 20 42
Screenshot 2020-08-05 at 11 20 32
Screenshot 2020-08-05 at 11 20 22
Screenshot 2020-08-05 at 11 21 48


FlorianCassayre:
Very nice addition.

I noticed that the two dropdowns had small differences (in alignment and layout). Moreover it is possible to open the "Sort" menu on top of "Filters" (the other way around works correctly).

image

In terms of UX it would be better if the filter menu did not reload each time the user makes a selection. I'm aware this would likely require a lot of changes on both mobile and pc. Regardless it would be good to automatically expand all the categories with at least one ticked element when the menu is opened.

Note that there is no way to know what the currently selected filter is. This could be solved by (for instance) adding an `active` class to the selected option.

src/lib/modules/SearchControls/overridden/SearchSortOrderElementMobile.js
```diff
@@ -0,0 +1,55 @@
...
+  sortChangeHandler = (onValueChange, currentSortOrder, options) => {
+    const value =
+      currentSortOrder === options[0].value
+        ? options[1].value
+        : options[0].value;
+    onValueChange(value);
+  };
+
+  render() {
+    const { currentSortOrder, options, onValueChange } = this.props;
```
zzacharo, line 20: Aren't the options fixed after received? If yes, we could avoid all this recomputation on every render and fill in the `this.buttons` property on the constructor. Then, you leave here only the `currentSortOrder` prop to control when you will need to rerender :)

kprzerwa:
> Very nice addition.
> 
> I noticed that the two dropdowns had small differences (in alignment and layout). Moreover it is possible to open the "Sort" menu on top of "Filters" (the other way around works correctly).
> 
> image
> 
> In terms of UX it would be better if the filter menu did not reload each time the user makes a selection. I'm aware this would likely require a lot of changes on both mobile and pc. Regardless it would be good to automatically expand all the categories with at least one ticked element when the menu is opened.
> 
> Note that there is no way to know what the currently selected filter is. This could be solved by (for instance) adding an `active` class to the selected option.

I've fixed all the problems, but for the selected filters I will start a new ticket as an improvement, since it will require bigger changes, good suggestion!

src/lib/modules/SearchControls/overridden/SearchSortOrderElementMobile.js
kprzerwa: done!

PR #153, opened 2020-08-05
forms: Fixes form bugs

### LINKED ISSUE #26: Forms: Fix empty input errors on submit
When submitting a form and the user starts entering info into a text field then removes the input, it will submit an empty string value and the schema validation will fail if it expects a string longer than 1. This also happens if the submitted value is an array and it's submitting an empty array but the schema validation expects at least one item.

A few possible solutions:
* Clean the submitted values
* Do not have min length validation for non-required fields

### LINKED ISSUE #74: add nested fields validator to <BaseForm>
Prevent the submit action if the nested required fields are not filled.

Proposed solution:
Add the generic handler which checks the fields and sets the button to disabled if fields not filled

jrcastro2:
* Fixes input errors on the fields
* Fixes bugs that caused errors
  when opening forms
* Fixes bug that caused the creation
  of 2 vendors
* Fixes bugs that caused error when
  submitting forms
* Fixes some more minor bugs
* closes #74
* closes #26

### DIFF
--- a/src/lib/forms/core/BaseForm/RecordSerializer.test.js
+++ b/src/lib/forms/core/BaseForm/RecordSerializer.test.js
@@ -1,0 +1,122 @@
+import { removeEmptyValues } from './RecordSerializer';
+
+describe('removeEmptyValues', () => {
+  it('should clean a simple object', () => {
+    const object = {
+      authors: [{ name: 'Joe', surname: '' }],
+      contributors: [{ identifiers: [] }],
+      roles: [],
+      version: 0,
+      cool: false,
+      creators: [null, undefined, {}],
+      description: '',
+    };
+
+    const expectedObject = {
+      authors: [{ name: 'Joe', surname: undefined }],
+      contributors: [{ identifiers: undefined }],
+      roles: undefined,
+      version: 0,
+      cool: false,
+      creators: undefined,
+      description: undefined,
+    };
+    removeEmptyValues(object);
+    expect(object).toEqual(expectedObject);
+  });
+
+  it('should clean a real example object', () => {
+    const object = {
+      authors: [
+        {
+          full_name: 'asd',
+          alternative_names: [],
+          affiliations: [],
+        },
+      ],
+      conference_info: {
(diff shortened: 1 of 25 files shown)


src/lib/forms/core/BaseForm/RecordSerializer.test.js
```diff
@@ -0,0 +1,122 @@
+import { removeEmptyValues } from './RecordSerializer';
+
+describe('removeEmptyValues', () => {
+  it('should clean a simple object', () => {
+    const object = {
+      authores: [{ name: 'Joe', surname: '' }],
```
zzacharo, line 6: typo :)

src/lib/forms/core/DateTimeFields/DateInputField.js
```diff
@@ -20,6 +25,15 @@ export class DateInputField extends React.Component {
           handleBlur={props.form.handleBlur}
           handleDateChange={(value, name) => {
             props.onChange(fromISO(value), name);
+            if (_isEmpty(value)) {
+              if (fieldPath.includes('.')) {
+                let splittedFieldPath = fieldPath.split('.');
```
zzacharo, line 30: I think you can deconstruct that to `let [keyPath, nestedKeyPath, ...parts] = fieldPath.split('.')`. Then, you can use the new variables below. I think it would make it easier to read. WDYT?

jrcastro2: Yes, I agree. Thanks!

src/lib/forms/core/DateTimeFields/DateInputField.js
```diff
@@ -20,6 +25,15 @@ export class DateInputField extends React.Component {
           handleBlur={props.form.handleBlur}
           handleDateChange={(value, name) => {
             props.onChange(fromISO(value), name);
+            if (_isEmpty(value)) {
```
kprzerwa, line 28: we could change the whole block to be a method of this component, for readability of the code

PR #158, opened 2020-08-13
Fix document form

### LINKED ISSUE #154: Document form - fixes
- [x] review the required fields and mark them with `*` on the field label
- [x] reorganise the fields - order by how important they are to fill in. The required ones should be on top
- [ ] source of the metadata - by default when the record is created via form the source is the user input (the field should be pre-filled with  "user" - please check the correct value in the vocabulary ), on edit it does not matter - it will take the initial value anyway

- [x] distinguish the minimal data section ( required fields ) from the rest of the sections of the form
- [ ] nested fields in the form should be distinguished. Now they look the same as the rest of the form and they blend in.
- [x] collapsable fields with Array type (f.e. keywords) should have one row already visible to fill in immediately.
Mockup:
document_form_mockup

jrcastro2:
* Reorganise the fields
* closes #154
### Document form
collapsed
### Non collapsed
Screenshot_2020-08-13 Screenshot

### DIFF
--- a/src/lib/forms/core/ObjectListField.js
+++ b/src/lib/forms/core/ObjectListField.js
@@ -42,7 +42,6 @@ export class ObjectListField extends Component {
 
     return (
       <Form.Field className="object-list-field" required>
-        <label>Authors</label>
         <List horizontal divided relaxed>
           {items.map((item, index) => (
             <List.Item
@@ -64,7 +63,7 @@ export class ObjectListField extends Component {
             onClick={() => this.onItemClick(null, items.length)}
           >
             <List.Content>
-              <Icon name="add" />
+              <Icon name="id card" />
               New author
             </List.Content>
           </List.Item>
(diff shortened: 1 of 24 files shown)


jrcastro2:
On the URLs, I feel like this is not enough space
image
But now it feels like is too much
image
WDYT?

FlorianCassayre:
In terms of usability the second option is definitely better. Visually, when both groups are empty, the first option looks better. However I don't think we can speculate on which one is going to take more space in practice, so overall I would (personally) still go with the second option.

src/lib/forms/core/ObjectListField.js
```diff
@@ -64,7 +63,7 @@ export class ObjectListField extends Component {
             onClick={() => this.onItemClick(null, items.length)}
           >
             <List.Content>
-              <Icon name="add" />
+              <Icon name="id card" />
```
kprzerwa, line 66: not blocking for now: as the field is meant to be the generic one (ObjectListfield), the icon ideally should be passed as a prop, to change it depending on the type of the object

src/lib/pages/backoffice/Document/DocumentForms/DocumentEditor/DocumentForm/DocumentAdditionalInfo.js
```diff
@@ -0,0 +1,41 @@
+import React, { Component } from 'react';
+import { Grid, Segment } from 'semantic-ui-react';
+import {
+  AlternativeAbstracts,
+  AlternativeIdentifiers,
+  AlternativeTitles,
+} from './components';
+
+export class DocumentAdditionalInfo extends Component {
```
kprzerwa, line 9: Nice I like the separation of the sections!

PR #161, opened 2020-08-13
form: fix AccordionField style

kprzerwa:
Screenshot 2020-08-13 at 17 12 39


kprzerwa:
because it is default semantic-ui accordion now
Screenshot 2020-08-14 at 11 23 15
and after we agreed the changing of the Icon to be visible in front of the text of the accordion, the angle down icon did not look good there

PR #176, opened 2020-08-31
Add static error pages

### LINKED ISSUE #21: frontsite: create static pages for http error codes
- [x] 401 Unauthorized
- [x] 403 Forbidden
- [x] 404 Not Found
- [x] 500 Internal Server Error

FlorianCassayre:
* pages for 401, 403 and 500
* also fixes a bug when displaying 404
* closes #21 

localhost_3000_sefes
localhost_3000_sefes (2)
localhost_3000_sefes (1)


ntarocco (commented):
Thanks @FlorianCassayre for working on this, and it look very nice!

My personal opinion is that it would be nice to add a more user friendly message if possible. I would to something like:

```
401 Unauthorized

Sorry, you are not authorized to see this page
```
(or something better, maybe a quick search on the web to see what is recommended)
Similar for `403`. I am actually wondering if we should put the HTTP status code 401/403 at all... is this of any help?

For the 500, I would go again for:
- a nice message: `Oups, something went wrong. The error has been sent to the administrators`... etc, with a much better and user friendly message
- can you investigate if it will be easy to expose in the REST APIs the `Sentry id` of the error and display it on the page if Sentry enabled? Like this we could show: `Please send this <id> to the support team to help them investigate the issue...

src/lib/pages/frontsite/Documents/DocumentDetails/DocumentDetails.js
```diff
@@ -160,7 +160,7 @@ class DocumentDetails extends Component {
         params: { documentPid },
       },
     } = this.props;
-    if (hasError && error.response.status === 404) {
+    if (hasError && error.response && error.response.status === 404) {
```
topless, line 163:
```suggestion
    if (hasError && _has(error.response.status) === 404) {
```
do not forget to import _has from lodash

topless:
Great job looks very good, I would agree with Nico's comment about being more explicit and the status codes, can be omitted for the end user.

As far as I can see, these components are extremely similar, they have an `icon`, an `http_code` and a `message`, the back to home button is the same everywhere.

Would it make sense instead of having explicit components for each error page, to make a more generic one, like ErrorPage component which will just take the aforementioned properties?

I didn't see their usage in this pr, so I might be missing something.

src/lib/components/Forbidden/Forbidden.js
```diff
@@ -0,0 +1,37 @@
...
+class Forbidden extends Component {
+  render() {
+    return (
+      <Overridable id="Forbidden.layout" {...this.props}>
+        <div className="frontsite">
+          <Grid
+            container
+            verticalAlign="middle"
+            textAlign="center"
+            className="not-found error-page"
```
topless, line 15: I am not able to found the declaration of the css classes `not-found`, `internal-server-error`, and `unauthorized`. If they are not used, I would suggest removing them, wdyt?

FlorianCassayre: This was meant for custom styling, instead of override the entire component (similar to what was done in `NotFound`). But I agree that `error-page` is more than enough for most purposes - and overriding is still possible in case of further modifications.

FlorianCassayre:
Thanks; I removed the error code and replaced it by a title + message. I extracted the layout as a separate overridable component, and created the four pages on top of it. The individual pages are also separately overridable. This component may also be used later in the backoffice (`isBackOffice` prop).

localhost_3000_inveniosoftware_react-invenio-app-ils

FlorianCassayre:
Uncaught exceptions and error should allow the server to include an `error_id` field in their body which would correspond to the sentry log id ([errors.py](https://github.com/inveniosoftware/invenio-app-ils/blob/master/invenio_app_ils/errors.py#L43)). Now I didn't manage to test this feature, even with `debug = False`.

src/lib/components/HttpErrors/HttpErrors.js
```diff
@@ -0,0 +1,64 @@
...
+      />
+    </Overridable>
+  );
+}
+
+function InternalServerErrorComponent(props) {
+  return (
+    <Overridable id="InternalServerError.layout" {...props}>
+      <HttpError
+        message="Internal Server Error"
```
topless, line 45: There is a mix here, between the title message and that is an interesting icon you got there :D

src/lib/components/HttpErrors/HttpErrors.js
```diff
@@ -0,0 +1,64 @@
...
+        title="Unauthorized"
+        message="You are not authorized to access this resource."
+        icon="ban"
+      />
+    </Overridable>
+  );
+}
+
+function ForbiddenComponent(props) {
+  return (
```
topless, line 18: Do you believe that giving the ability to override the values with props would make sense? Something like 

```
const {
        title="Access Forbidden"
        message="You do not have the sufficient privileges to access this resource."
        icon="ban"
} = this.props
```
and then 
```
      <HttpError
        title={title}
        message={message}
        icon={icon}
      />
```
So instead of redefining a component we could do
`<Forbidden title="My special forbidden title for this case">`

src/lib/components/HttpErrors/HttpError.js
```diff
@@ -0,0 +1,50 @@
...
+            container
+            verticalAlign="middle"
+            textAlign="center"
+            className="error-page"
+          >
+            <Grid.Column>
+              <Icon name={icon} size="massive" />
+              <h1>{title}</h1>
+              <h3>{message}</h3>
+              {isBackOffice || (
```
topless, line 23: I am not sure what we are trying to achieve here, shouldn't the link `Back home` be displayed in every case? On a side note I believe for conditional rendering we need `&&` not `||`.

FlorianCassayre: I changed it because it didn't make sense to have a link "Back home" when the page was displayed in the backoffice. I can also keep the button and change the link to the backoffice's homepage.

src/lib/components/HttpErrors/HttpErrors.js
FlorianCassayre: Yes it makes sense, good suggestion!

topless (approved):

ntarocco:
> Uncaught exceptions and error should allow the server to include an `error_id` field in their body which would correspond to the sentry log id ([errors.py](https://github.com/inveniosoftware/invenio-app-ils/blob/master/invenio_app_ils/errors.py#L43)). Now I didn't manage to test this feature, even with `debug = False`.

@FlorianCassayre you are totally right, I did not give enough details. [Here](https://github.com/inveniosoftware/invenio-rest/blob/580e7da13d2f5b085faad436446a44f6a8f7374c/invenio_rest/views.py#L39) you have the code, you can set it globally in `g` and test it. Would you mind it?
Then we could have a 500 page with a text similar to:
```
title = 'Something went wrong',
message = 'An unexpected error has occurred. Please try again or contact our support at <email in config> specifying the error code <sentry id>),
```
The last part with the error code should appear only if returned by the webservice...
WDYT?

See: https://duckduckgo.com/?t=ffab&q=Something+went+wrong+error+message&iax=images&ia=images

FlorianCassayre:
:+1: Planned for this afternoon.

FlorianCassayre:
Here are my latest thoughts:

* When a 500 error occurs on any endpoint, we would intercept it on [base.js:23](https://github.com/inveniosoftware/react-invenio-app-ils/blob/master/src/lib/api/base.js#L23), and trigger a (new) action
* A component placed at the level of [BackOffice.js:13](https://github.com/inveniosoftware/react-invenio-app-ils/blob/master/src/lib/routes/backoffice/BackOffice.js#L13) would have a conditional display of the main content depending on the value of this action. In case of an error it would replace the pane content by the `500` error page. Similar pattern on the frontsite
* That way we ensure that if something is wrong the interactive part of the interface disappears. But most importantly we don't need to implement fine grained logic for each page as it would be handled globally
* This could be extended to HTTP errors other than `500`; I can think of `403`, `404` and `429`

It would essentially look like the following with the corresponding error page displayed (again, same for the frontsite):

image

kprzerwa (approved):

PR #179, opened 2020-08-31
Add keywords and tags to series

FlorianCassayre:
* covers frontsite & backoffice
* refactored code into components
* depends on inveniosoftware/invenio-app-ils#930
* closes inveniosoftware/invenio-app-ils#917

### DIFF
--- a/src/lib/pages/backoffice/Series/SeriesDetails/SeriesMetadata/SeriesMetadata.js
+++ b/src/lib/pages/backoffice/Series/SeriesDetails/SeriesMetadata/SeriesMetadata.js
@@ -8,6 +8,8 @@ import _isEmpty from 'lodash/isEmpty';
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 import { Container, Divider } from 'semantic-ui-react';
+import LiteratureTags from '@modules/Literature/LiteratureTags';
+import LiteratureKeywords from '@modules/Literature/LiteratureKeywords';
 
 export default class SeriesMetadata extends Component {
   prepareData = () => {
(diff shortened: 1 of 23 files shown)


src/lib/modules/Document/DocumentKeywords.js
```diff
@@ -0,0 +1,47 @@
...
+              .map(keyword => displayKeyword(keyword.value, keyword.source))
+              .join(separator)
+          : noneMessage}
+      </>
+    </Overridable>
+  );
+};
+
+DocumentKeywords.propTypes = {
+  metadata: PropTypes.object.isRequired,
```
ntarocco, line 37: I would prefer to pass only the keywords list if possible, instead of the entire metadata obj.
```suggestion
  keywords: PropTypes....
```

src/lib/pages/backoffice/Series/SeriesDetails/SeriesMetadata/SeriesMetadata.js
```diff
@@ -29,6 +31,20 @@ export default class SeriesMetadata extends Component {
...
+        name: 'Keywords',
+        value: <DocumentKeywords keywords={seriesDetails.keywords} />,
+      },
+      {
+        name: 'Tags',
+        value: (
+          <DocumentTags
+            isBackOffice
+            size="mini"
+            metadata={seriesDetails.metadata}
```
ntarocco, line 44: It was not you creating this, but would it be possible to fix also this one and pass only the tags instead?

FlorianCassayre: Sure thing, there is at least one more component using this pattern (that's where I took the inspiration from). Will fix all of them.

FlorianCassayre:
Hopefully I refactored everything correctly (I double checked but I still might have missed something).

kprzerwa (approved):
not a blocker: since `DocumentKeywords` and `DocumentTags` are used with the series rectype too, the name should follow the convention of `LiteratureTags`, `LiteratureKeywords`.

Issue #172, opened 2020-08-14
UI testing: issues


FlorianCassayre:
## Backoffice

  - Any form that uses a "search document/location" component:
    - [x] When no result is found the dropdown stays visible, even when the field loses focus. Actually the only way to get rid of the message is to make a selection. (fixed by #217)
    - [x] The auto-completion does not work as the user intends it to: it expects full keywords rather than just prefixes  (fixed by #217)
  - Document/Series/Item/... forms:
    - [x] No error message is displayed if the server is down (see also inveniosoftware/invenio-app-ils#855) (fixed by #220)
    - [x] ~The error tooltip on fields is displayed on top, messing the alignment~ (see #169)
    - [x] ~The price field does not match the server schema~ (see #155)
    - [x] The "Navigation" menu does not have an initial value before scrolling (fixed by #180)
    - [x] Some elements define a `"tabindex"="0"` and are initially focused, but actually appear in the middle of the page (fixed by #180)
    - [x] Frontend error when viewing the page from a mobile device (fixed by #180)
    - [x] The cross icon (keywords, authors, ...) appears on top of everything else, including an expanded dropdown(fixed by #216)
  - Search:
    - [x] Search result exportation doesn't work (fixed by #180)
  - Document search:
    - [x] ~When clicking on a tag (Drama, Thriller, etc.), the url is updated but nothing happens unless the page is reloaded~ (see #196)
    - [x] ~Start with an empty search, click on a tag, and go back a page: the search provides an empty result set (again, unless if the page is reloaded)~ (see #196)
  - Document:
    - [x] In the "Request loan for a patron" modal, the "Request" button is not disabled when no patron is provided (fixed by #180)
    - [x] When in document details, the button `Request loan for Patron` in `DocumentActionMenu`, throws an error: `locationPid` is marked as required in `LocationDatePicker`, but its value is `null`. The input for the date is disabled until we select a patron with a `locationPid`, maybe we can prevent rendering the calendar until we have a `locationPid`(fixed by #216)
  - Loans:
    - [x] Frontsite error after clicking on the "Replace" button ([trace](https://user-images.githubusercontent.com/9027075/90262901-361eeb00-de4f-11ea-9ea3-1ea01b6e98be.png)) (fixed by #180)
    - [x] ~"Send return reminder", is not working for overdue loans (it returns a 400 saying "This loan is not overdue")~ (see #204)
  - Item:
[...]

PR #180, opened 2020-09-01
Minor bugfixes

### LINKED ISSUE #135: literature: cover image displays previous state
When navigating from one literature page to another, the cover image of the last viewed literature is displayed while the image of the current one is loading.
What should happen: all the literature pages (and all pages in general) should be isolated from each other.

FlorianCassayre:
* addresses #172
* DocumentAuthors component raising exception
* "checkout" button in loan requests not disabled during loading
* LiteratureCover flickering on page change (closes #135)
* hide navigation buttons in patron loans on empty result set
* BorrowingRequestPatronLoan state not reset
* ESSelectorLoanRequest request button not disabled
* SearchSortOrderElementMobile uninitialized attribute
* ExportSearchResults not starting from the first page
* ScrollingMenu first element not selected due to typo
* HitsSearch & ESSelector toggleable initial focus, disabled by default
* LoanMetadata invalid dates displayed

### DIFF
--- a/src/lib/components/backoffice/ExportSearchResults/ExportSearchResults.js
+++ b/src/lib/components/backoffice/ExportSearchResults/ExportSearchResults.js
@@ -105,6 +105,7 @@ class ExportSearchResultsWithState extends Component {
 
     // append the `format` param
     const params = Qs.parse(queryString);
+    params['page'] = 1; // Start from the first page
     params['format'] = format;
     const args = Qs.stringify(params);
 
--- a/src/lib/components/backoffice/buttons/ScrollingMenu/ScrollingMenu.js
+++ b/src/lib/components/backoffice/buttons/ScrollingMenu/ScrollingMenu.js
@@ -5,7 +5,7 @@ import { Menu } from 'semantic-ui-react';
 export default class ScrollingMenuItem extends Component {
   constructor(props) {
     super(props);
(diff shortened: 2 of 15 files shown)


src/lib/components/backoffice/ExportSearchResults/ExportSearchResults.js
```diff
@@ -105,6 +105,7 @@ class ExportSearchResultsWithState extends Component {
 
     // append the `format` param
     const params = Qs.parse(queryString);
+    params['page'] = 1; // Start from the first page
```
kprzerwa, line 108: how will it behave with other pages?

kprzerwa (approved):

src/lib/components/backoffice/ExportSearchResults/ExportSearchResults.js
FlorianCassayre: The button has the same effect, regardless the currently selected page. Namely it exports the results of the search without taking pagination into account. I assumed this was the intended semantic of that button since it's said to return the "first 10,000 records".
Did I miss something?

kprzerwa: I need to investigate this further to make sure, I will merge this and start a ticket

PR #181, opened 2020-09-01
oauth: add enabled attribute

ntarocco:


src/lib/config/index.js
```diff
@@ -1,20 +1,18 @@
...
   constructor() {
-    this.setValue(defaultConfig);
+    this.APP = APP_CONFIG;
+    Object.assign(this, { ...RECORDS_CONFIG });
   }
 
-  setValue = newConfig => {
-    _assign(this, _merge({}, this, newConfig));
+  update = newConfig => {
+    Object.assign(this, _merge(this, newConfig));
```
topless, line 15: Would it make sense to `Object.freeze()` when we call the update to prevent further changes?

topless (approved):

ntarocco:
@topless can you please re-check it?

PR #187, opened 2020-09-08
 fronsite: Updates error message in login page

### LINKED ISSUE #145: Login: fix error message and OAuth
There are 2 issues in the Login.js component:

- [x] Notifications (success or error) are shown 2 times, probably because the `<Notifications` cmp is added.
- [ ] Notifications are outside the background.
- [x] There should be no notifications in the frontsite, replace them with a box in the login page
- [x] It is not possible to override the OAuth config and removing existing OAuth login (e.g. `GitHub`): probably when merging overridden config, we should use `assign` instead of `merge`? To be checked.

See screenshot:

login

jrcastro2:
* Deletes the notifications with the error
  messages and displays this messages in
  a box.
* closes #145

image
------------------
image


KonstantinaStoikou (approved):

ntarocco (approved):
Looks good to me, thanks a lot for fixing this!

Issue #191, opened 2020-09-11
Fix styling of ESSelector Selected values


kprzerwa:
Elasticsearch search components: is it possible to improve the display when a value is selected? 
it is currently not very visible that the selected value belongs to the ESSelector field, it is displayed below with a x to remove it.

PR #193, opened 2020-09-11
UX improvements

### LINKED ISSUE #25: Frontsite: cannot open documents in a new tab
In the frontsite, right-click -> open in a new tab on any document cannot be done (no real links).
It looks like it is happening in the homepage and search results.

FlorianCassayre:
* fixed dynamic links (closes #25)
* fixed some react related errors
* improved ESSelector component (addresses #191)
* refactoring (using the dedicated icon components when possible)

image

image

### DIFF
--- a/src/lib/components/backoffice/MetadataTable/MetadataTable.js
+++ b/src/lib/components/backoffice/MetadataTable/MetadataTable.js
@@ -8,7 +8,9 @@ export default class MetadataTable extends Component {
 
     return rows.map(row => (
       <Table.Row key={row.name}>
-        <Table.Cell width={labelWidth}>{row.name}</Table.Cell>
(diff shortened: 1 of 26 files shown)


FlorianCassayre:
The selection input can be improved even further by converting the `HitsSearch` component into a `Dropdown`. Would work for both single and multiple selections.

ntarocco (approved):
Thanks for these fixes!
A good UI would be that we have:

**Searching:**

searching

**Selected:**

selected

FlorianCassayre:
Yep, I was thinking the same. Currently only the `Dropdown` component is able to display labels inside the input, but it has its own search mechanism and I'm not sure if we can combine it with the `Search` component.

PR #194, opened 2020-09-11
Backoffice: Fixes patron page checkout

### LINKED ISSUE #79: Backoffice patron page: re-design the search box
- [x] The text "No barcode provided" is too visible. The issue is that it is a bit misleading to see that message after a checking/checkout. The box should be re-designed.
- [x] The search box content should be reset when leaving the page. Currently it keeps the previous value when going away and coming back.
- [x] When we insert a barcode typing it and press enter an error ocurrs. Needs to be fixed.

jrcastro2:
* Removes misleading box
* Now it cleans the results when
  leaving the page
* Fixes bug that prevents from searching
  by pressing enter
* Adds a loading state for the search
* closes #79

### DIFF
--- a/src/lib/pages/backoffice/Patron/PatronDetails/ItemsSearch/ItemsResultsList/ItemsResultsList.js
+++ b/src/lib/pages/backoffice/Patron/PatronDetails/ItemsSearch/ItemsResultsList/ItemsResultsList.js
@@ -6,7 +6,7 @@ import { BackOfficeRoutes } from '@routes/urls';
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 import { Link } from 'react-router-dom';
-import { Button, Header, Icon, Modal, Segment, Popup } from 'semantic-ui-react';
+import { Button, Header, Icon, Modal, Message, Popup } from 'semantic-ui-react';
 import _isEmpty from 'lodash/isEmpty';
 
 export default class ItemsResultsList extends Component {
@@ -138,7 +138,12 @@ export default class ItemsResultsList extends Component {
   };
 
   render() {
-    const { results, checkoutIsLoading, clearSearchQuery } = this.props;
+    const {
+      results,
+      checkoutIsLoading,
+      executedSearch,
+      isLoading,
+    } = this.props;
     const columns = [
       {
         title: 'Barcode',
@@ -155,7 +160,9 @@ export default class ItemsResultsList extends Component {
       },
     ];
 
-    return !_isEmpty(results.hits) ? (
+    return isLoading && executedSearch ? (
+      <Loader isLoading={isLoading} />
+    ) : !_isEmpty(results.hits) ? (
       <Loader isLoading={checkoutIsLoading}>
         <p>Found {results.hits.length} item(s).</p>
         <ResultsTable
@@ -165,19 +172,13 @@ export default class ItemsResultsList extends Component {
           name="items"
         />
       </Loader>
-    ) : (
-      <Segment placeholder textAlign="center">
-        <Header icon>
-          <Icon name="search" />
-          Found no items matching this barcode.
-        </Header>
-        <Segment.Inline>
-          <Button primary onClick={() => clearSearchQuery()}>
-            Clear search phrase
-          </Button>
-        </Segment.Inline>
-      </Segment>
-    );
+    ) : executedSearch ? (
+      <Message
+        error
+        header="Found no physical copies for this barcode."
+        content="Please try again"
+      />
+    ) : null;
   }
 }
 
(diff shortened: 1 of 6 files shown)


jrcastro2:
### Before
image

### After
image
------------------------
image

FlorianCassayre (commented):
@jrcastro2 If you paste the same code twice, it dispatches an error notification on the second time.

src/lib/pages/backoffice/Patron/PatronDetails/ItemsSearch/ItemsSearch.js
```diff
@@ -16,27 +16,29 @@ export default class ItemsSearch extends Component {
...
-    return fetchItems(queryString);
+    this.setState({ prevSearchQuery: queryString, executedSearch: true });
+    fetchItems(queryString);
   };
 
   onPasteHandler = async event => {
     const { checkoutItem, patronDetails } = this.props;
+    const { prevSearchQuery } = this.state;
     let queryString = event.clipboardData.getData('Text');
+    let doublePaste = prevSearchQuery === queryString;
```
ntarocco, line 39:
```suggestion
    const sameQueryString = prevSearchQuery === queryString;
```
I would change `let` to `const` whenever possible and also the name of the var: it could be that there was no pasting event eventually, but typing the same query string

src/lib/pages/backoffice/Patron/PatronDetails/ItemsSearch/ItemsSearch.js
```diff
@@ -53,15 +55,16 @@ export default class ItemsSearch extends Component {
           type: recordToPidType(hits[0]),
           value: hits[0].metadata.pid,
         };
-        checkoutItem(documentPid, itemPid, patronDetails.user_pid, true);
+        await checkoutItem(documentPid, itemPid, patronDetails.user_pid, true);
```
ntarocco, line 58: can you double check that there is a try/catch for these 2 async/await, just to be sure we catch the error?

ntarocco (approved):

src/lib/pages/backoffice/Patron/PatronDetails/ItemsSearch/ItemsResultsList/ItemsResultsList.js
```diff
@@ -155,7 +160,11 @@ export default class ItemsResultsList extends Component {
       },
     ];
 
-    return !_isEmpty(results.hits) ? (
+    return isLoading && executedSearch ? (
+      <Loader isLoading={isLoading}>
+        <this.renderAll />
```
ntarocco, line 165: I am not sure this `renderAll` is correct, I cannot find it. What was the expected behaviour?

src/lib/pages/backoffice/Patron/PatronDetails/ItemsSearch/ItemsResultsList/ItemsResultsList.js
```diff
@@ -165,30 +174,27 @@ export default class ItemsResultsList extends Component {
...
-          <Button primary onClick={() => clearSearchQuery()}>
-            Clear search phrase
-          </Button>
-        </Segment.Inline>
-      </Segment>
-    );
+    ) : executedSearch ? (
+      <Message
+        error
+        header="Found no items matching this barcode."
```
ntarocco, line 180:
```suggestion
        header="Found no physical copies for this barcode."
```

ntarocco:
Last question, can you try to use the truncate `prop` for the `<LiteratureTitle` in these boxes?

https___localhost_3000_backoffice_patrons_2

src/lib/pages/backoffice/Patron/PatronDetails/ItemsSearch/ItemsResultsList/ItemsResultsList.js
jrcastro2: No it's not correct, thanks for noticing. I probably forget to delete it from a copy paste, the expected behaviour is that if it is executing the searching query avoid to display the `Found no items...` and display the `loader` even if it only lasts some miliseconds before displaying the result list (if there are elements)

jrcastro2:
> Last question, can you try to use the truncate `prop` for the `<LiteratureTitle` in these boxes?
> 
> https___localhost_3000_backoffice_patrons_2

In order to get the truncate to fix this problem we need to add the width property, otherwise if the title is too long it will get bugged anyway.
Truncate without width
image
Truncate with width
image

ntarocco (approved):

ntarocco:
Thanks a lot for the fixes!!!

ntarocco (approved):

PR #201, opened 2020-09-18
ItemDetails: move circulation label to title

### LINKED ISSUE #83: Backoffice: labels should be more consistent
From the library:
```
- If on loan, the information appears "on loan" at the top of the page. If available, the information appears below "Physical copy on shelf".
- The Item status displayed does not always correspond to the real status. Ex. The item status shows: "Physical copy on shelf", but there is a loan on it. | The loan status doesn't display the same way depending if the item is on loan or not.. It would be preferable to display the same way at the top of the page. "On loan" "On shelf".
```

In item details page, when the item is on loan, the label ON-LOAN is appearing.
When the item is not on loan, nothing appears.
The divider shows different labels (STATUS/LOAN).

Improve consistency.

KonstantinaStoikou:
Closes #83 

Screenshot 2020-09-18 at 10 18 37

Screenshot 2020-09-18 at 10 19 28

### DIFF
--- a/src/lib/pages/backoffice/Item/ItemDetails/ItemHeader.js
+++ b/src/lib/pages/backoffice/Item/ItemDetails/ItemHeader.js
@@ -8,9 +8,41 @@ import { BackOfficeRoutes } from '@routes/urls';
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 import { Link } from 'react-router-dom';
-import { Header } from 'semantic-ui-react';
+import { Header, Label } from 'semantic-ui-react';
+import { invenioConfig } from '@config';
 
 export class ItemHeader extends Component {
+  statusLabel = () => {
+    const { data } = this.props;
+    let status = { value: 'on-shelf', text: 'On shelf' };
+    if (
+      invenioConfig.ITEMS.canCirculateStatuses.includes(data.metadata.status)
+    ) {
+      if (
+        invenioConfig.CIRCULATION.loanActiveStates.includes(
+          data.metadata.circulation.state
+        )
+      ) {
+        status = invenioConfig.CIRCULATION.statuses.find(
+          x => x.value === data.metadata.circulation.state
+        );
+      }
+    } else {
+      status = invenioConfig.ITEMS.statuses.find(
+        x => x.value === data.metadata.status
+      );
+    }
+    const lowercaseDashed = value => {
+      value = value.toLowerCase().replace(/_/g, '-');
+      return value;
+    };
+    return (
+      <Label className={`item-status-${lowercaseDashed(status.value)}`}>
+        {status.text}
+      </Label>
+    );
+  };
+
   render() {
     const { data } = this.props;
     const recordInfo = (
@@ -40,6 +72,7 @@ export class ItemHeader extends Component {
               showOnlyTitle
               truncate
             />
+            {this.statusLabel()}
           </>
         }
         subTitle={
--- a/src/lib/config/defaultConfig.js
+++ b/src/lib/config/defaultConfig.js
@@ -159,15 +159,30 @@ export const RECORDS_CONFIG = {
     loanRequestStates: ['PENDING'],
     requestDuration: 60,
     statuses: [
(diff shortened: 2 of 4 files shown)


src/lib/pages/backoffice/Item/ItemDetails/ItemHeader.js
```diff
@@ -8,11 +8,29 @@ import { BackOfficeRoutes } from '@routes/urls';
...
+  };
+
   render() {
     const { data } = this.props;
+
+    const statusLabel = circulationState => {
+      return this.hasActiveLoan(circulationState) ? (
+        <Label color="purple">ON LOAN</Label>
+      ) : (
+        <Label color="blue">ON SHELF</Label>
```
ntarocco, line 30: what about if it on binding/missing?

src/lib/pages/backoffice/Item/ItemDetails/ItemHeader.js
```diff
@@ -8,9 +8,37 @@ import { BackOfficeRoutes } from '@routes/urls';
...
+        status = invenioConfig.CIRCULATION.statuses.find(
+          x => x.value === data.metadata.circulation.state
+        );
+      }
+    } else {
+      status = invenioConfig.ITEMS.statuses.find(
+        x => x.value === data.metadata.status
+      );
+    }
+    return <Label className={status.styleClass}>{status.text}</Label>;
```
ntarocco, line 39: after discussions with @kprzerwa too, my thinking was that I would not add a configuration `styleClass` in the `config.js` that is for a specific component (the label), because if in the future we have to add a new component that will need different classes, then we will start adding them to the config.

I would do instead:
```
    return <Label className=`item-status{lowercaseDashed(status.value)}`>{status.text}</Label>;
```
so that you will have in the end:
```
<Label className='item-status-for-reference-only'>For reference only</Label>;
<Label className='item-status-in-binding'>In binding</Label>;
```

and then only create all the CSS classes as you have already done:
```
.item-status-for-reference-only.label {
    background-color: #6435c9 !important;
    border-color: #6435c9 !important;
    color: #ffffff !important;
  }
```

What do you think?

ntarocco (approved):

src/lib/pages/backoffice/Item/ItemDetails/ItemHeader.js
```diff
@@ -8,9 +8,45 @@ import { BackOfficeRoutes } from '@routes/urls';
...
+    let status = {};
+    if (
+      invenioConfig.ITEMS.canCirculateStatuses.includes(data.metadata.status)
+    ) {
+      if (
+        !invenioConfig.CIRCULATION.loanActiveStates.includes(
+          data.metadata.circulation.state
+        )
+      ) {
+        status = {};
```
topless, line 26: I don't think this is needed

src/lib/pages/backoffice/Item/ItemDetails/ItemHeader.js
```diff
@@ -8,9 +8,45 @@ import { BackOfficeRoutes } from '@routes/urls';
...
+        status.value = 'on-shelf';
+        status.text = 'On shelf';
```
topless, line 28:
```suggestion
        status = { value: 'on-shelf', text: 'On shelf'};
```

src/lib/pages/backoffice/Item/ItemDetails/ItemHeader.js
```diff
@@ -8,9 +8,45 @@ import { BackOfficeRoutes } from '@routes/urls';
...
 import React, { Component } from 'react';
 import { Link } from 'react-router-dom';
-import { Header } from 'semantic-ui-react';
+import { Header, Label } from 'semantic-ui-react';
+import { invenioConfig } from '@config';
 
 export class ItemHeader extends Component {
+  statusLabel = () => {
+    const { data } = this.props;
+    let status = {};
```
topless, line 17: Since its is expected to have a `value` and a `text` properties, may would make sense to initialize them here.

topless (approved):

PR #208, opened 2020-09-29
frontsite: publication information

kprzerwa:
Screenshot 2020-09-29 at 10 20 30

### DIFF
--- a/src/lib/modules/Document/DocumentPublicationInfo.js
+++ b/src/lib/modules/Document/DocumentPublicationInfo.js
@@ -10,8 +10,8 @@ export class DocumentPublicationInfo extends Component {
     if (_isEmpty(publications)) {
       return (
         <InfoMessage
-          title="No publication information available."
-          message="This literature has not been published in any journal."
+          title="Journal publication"
+          message="No publication information for this literature."
         />
       );
     }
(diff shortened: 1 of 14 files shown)


src/lib/modules/Document/DocumentPublicationInfo.js
```diff
@@ -0,0 +1,71 @@
...
+import _isEmpty from 'lodash/isEmpty';
+
+export class DocumentPublicationInfo extends Component {
+  render() {
+    const { publications, documentType } = this.props;
+    if (_isEmpty(publications)) {
+      return (
+        <InfoMessage
+          title="No publication information available."
+          message="This literature has not been published in any journal."
```
ntarocco, line 14: if it is a book, maybe we don't want to show this? Shall we display it or now based on the `document_type`?

src/lib/modules/Document/DocumentStats.js
```diff
@@ -27,7 +27,7 @@ export class DocumentStats extends Component {
     const pidType = recordToPidType(document);
     try {
       this.cancellableFetchStats = withCancel(
-        statsApi.recordStats(pidType, document.pid)
+        statsApi.recordStats(pidType, document.metadata.pid)
```
ntarocco, line 30: how was it working before? 😱

src/lib/pages/frontsite/Documents/DocumentDetails/DocumentMetadata/DocumentMetadataTabs/DocumentMetadataTabs.js
```diff
@@ -50,7 +51,12 @@ class DocumentMetadataTabs extends Component {
       },
       {
         menuItem: 'Publications',
-        render: () => <Tab.Pane>We wait for the schema!</Tab.Pane>,
```
ntarocco, line 53: 💯 👍 🥳

src/lib/modules/Document/DocumentPublicationInfo.js
```diff
@@ -0,0 +1,71 @@
...
+      );
+    }
+    return (
+      <>
+        <Divider horizontal>Publication information</Divider>
+        This {documentType.toLowerCase()} was published in:
+        {publications.map((publication, idx) => (
+          <>
+            {idx > 0 && <Divider horizontal />}
+            <Table definition key={publication.journal_title}>
```
topless, line 25: Does it make sense to display them all, even if there is no information about it? So for example if the `pages` information is not available we should entirely omit pages table row.

src/lib/pages/frontsite/Documents/DocumentDetails/DocumentMetadata/DocumentMetadataTabs/DocumentMetadataTabs.js
```diff
@@ -120,6 +126,7 @@ class DocumentMetadataTabs extends Component {
         panes={this.renderTabPanes()}
         onTabChange={this.onTabChange}
         id="document-metadata-tabs"
+        className="document-metadata-tabs"
```
topless, line 129: Here we are using the same value for `id` which is meant to be unique among the dom elements, but also for class name which is meant to be shared among elements that belong to the same class. Personally I would go for something more explicit for the `id`.

src/lib/modules/Document/DocumentPublicationInfo.js
kprzerwa: From what I remember we decided to display everything, if it is not there to put the information like this one, in order not to have if statements everywhere

src/lib/modules/Document/DocumentStats.js
kprzerwa: because of this: https://github.com/inveniosoftware/react-invenio-app-ils/pull/208/files#diff-34d9d992c176e0090a1cd696e93660a8R181
I prefer to keep the structure of the original object though, it is more readable IMO

src/lib/modules/Document/DocumentPublicationInfo.js
ntarocco: Agree, but I guess it will be a bit weird in this case? A book "published" in a journal?
Maybe we can put a different message?
```
title="Journal publication"
message="No publication information for this literature"
```
or similar...

kprzerwa: 👍  it sounds better

src/lib/modules/Document/DocumentPublicationInfo.js
kprzerwa: fixed

src/lib/pages/frontsite/Documents/DocumentDetails/DocumentMetadata/DocumentMetadataTabs/DocumentMetadataTabs.js
kprzerwa: for me the `#document-metadata-tabs` and `.document-metadata-tabs` are very different, but I guess you mean for search in the code purposes. Fixed

src/semantic-ui/site/collections/table.overrides
```diff
@@ -38,6 +38,13 @@ Table Overrides - REACT-INVENIO-APP-ILS
 
 @{fs-parent-selector},
 @{bo-parent-selector} {
+
+  .document-metadata-tabs{
```
topless, line 42: It seems the `.overrides` files are not formatted in some cases there is a following space after names other times not. A quick fix is to open them as `Less` file type in VsCode, then prettier will do the rest.

topless (approved):

PR #209, opened 2020-10-01
Backoffice: improve vocabulary fetch error handling

### LINKED ISSUE #12: Vocabulary field error handling
When a vocabulary field fails to fetch data, it displays the following value:
`options: [{ key: '', value: '', text: 'Failed to load currencies.' }],`

The user won't easily see that there is an error in the field, because if it not highlighted in red.

This should be part of the vocabularies refactoring, when the vocabularies module will be defined.

FlorianCassayre:
* systematic and coherent errors for vocabulary fields
* disable empty option on error in multiple select
* fix and refactor price field
* closes #12 

image

image

image

(it was not possible to have a tooltip for this special component - therefore if no option is selected the error message is displayed as text)


ntarocco (approved):
Nice refactoring, thanks a lot!

kprzerwa (approved):

PR #210, opened 2020-10-01
backoffice: Fixes date validation

### LINKED ISSUE #174: Date validation fails
- [x] Fix in the borrowing request form the due date validation (it displays `Not a valid date`, even if it it should be valid )
- [x] If you submit a borrowing request form with the filed `debit date` filled but without filling the `payment` field then you get a warning since this field is required, if you try to submit again you will get an error.
Update:
- [x] Remove all the serializers functions that transform the date `string` fetched from ES to a `dateTime` object
image

jrcastro2:
* Fixes "due date" field in the borrowing request
  form
* Fixes date validation
* closes #174


topless (approved):

FlorianCassayre (approved):

src/lib/api/date.js
```diff
@@ -25,7 +25,11 @@ export const toISO = date => {
  *  @returns {String}
  */
 export const toISODate = date => {
-  return date.toISODate();
+  if (date instanceof DateTime) {
```
ntarocco, line 28: Let's discuss IRL tomorrow! ;)

src/lib/api/date.js
```diff
@@ -25,7 +25,7 @@ export const toISO = date => {
  *  @returns {String}
  */
 export const toISODate = date => {
-  return date.toISODate();
+  return date instanceof DateTime ? date.toISODate() : date;
```
ntarocco, line 28: Question for everyone @jrcastro2 @topless @FlorianCassayre @KonstantinaStoikou @kprzerwa, after the discussion we had, does it make sense to keep this util module? Wondering if it is not just easier to use `luxon` directly around instead of "wrapping" it here...

src/lib/api/documents/serializer.js
```diff
@@ -1,20 +1,12 @@
-import { fromISO, toISODate } from '@api/date';
+import { toISODate } from '@api/date';
 import { recordResponseSerializer } from '@api/utils';
 import _get from 'lodash/get';
 import _set from 'lodash/set';
 
 const DocumentSerializers = {
   DATE_FIELDS: ['circulation.next_available_date'],
   responseSerializer: function(hit) {
-    const result = recordResponseSerializer(hit, function(metadata) {
```
ntarocco, line 9: You can remove the entire serializer and DATE_FIELDS, the entire file, I believe...
Same for all of them below...

ntarocco (commented):
Double check of `DATE_FIELDS` or the usage of `fromISO` inside the `api` folder, there might be more...

src/lib/api/documents/serializer.js
kprzerwa: just be careful on things not working after, especially we might have places where we use `document.pid` instead of `document.metadata.pid` because of these serializers

src/lib/api/date.js
kprzerwa: I'm OK with removing it, however it means a bit of time spent on refactoring (please remove if it is not a time eating monster)

src/lib/api/documents/serializer.js
jrcastro2: I am not sure I am following ... As I understood in the discussion we had last week we agreed to remove the deserialization of the data when fetching it from ES/DB but we need to keep the serialization before the request, so we make sure that we send a `String` instead of `DateTime` object, since we store dates as `String` and if we enter a date in any of the forms we have, we create it as a `DateTime`. If we remove the entire file we would be sending `DateTime` objects and we would get an error. Maybe I am missing something ? 

If possible, I would like to discuss this IRL.

P.S: The `documents/serializer.js` file could be removed since we are not using it to serialize the dates (there are no dates).

ntarocco (approved):
Great thanks a lot!!! 💯 
Since this one is a bit "delicate", let's merge it and see if we apply any fix later one!

PR #137, opened 2020-07-28
search: sort by title

FlorianCassayre:
* added config fields to allow sorting the catalogue by titles (available in both frontsite and backoffice)
* depends on inveniosoftware/invenio-app-ils#923
* closes #111

PR #214, opened 2020-10-01
ReactSearchKit: pass initial query state [DEPENDS ON inveniosoftware/react-searchkit#137]

### LINKED ISSUE #196: Search: issues with history
- [x] On the frontsite: when performing a global search from the homepage, the pagination does not show any page to be selected. If you then click on "1" to select the first page, the result changes

 [PR](https://github.com/inveniosoftware/react-invenio-app-ils/pull/211) already for these:
- [x] On the same page: when clicking on a tag (Drama, Thriller, etc.), the url is updated but nothing happens.  
- [x] On the same page: when you have chosen some filters and you click the `Books` sidebar link it doesn't clear the filters (same happens for all the other searches that have filters and sidebar links).
If you reload the page (<kbd>F5</kbd>) all the above issues get resolved.
- [x] On the document search page: start with an empty search, click on a tag, and go back a page: the search provides an empty result set. (fixed because of unclickable tags)

### LINKED ISSUE #207: Search fails over 10000k results
## Describe the bug
when navigating in the search to page 700 (15 results per page = 10500) the search interface crashes:

Screenshot 2020-09-29 at 08 58 53
Screenshot 2020-09-29 at 08 59 05
Screenshot 2020-09-29 at 08 59 18

see also: https://github.com/inveniosoftware/react-invenio-app-ils/issues/189

KonstantinaStoikou:
Closes #196
Removes sortOrder dropdown from searches (and merges it with sortBy dropdown)  (except for searches that have many date options, e.g. Loans)
Closes #207


FlorianCassayre (commented):
I'm afraid it only addresses the issue partly; for instance consider clicking on one of the three button present on the frontsite homepage. Moreover, this was not mentioned in the original issue but it seems **all** the search forms are affected by this problem (backoffice in particular).

KonstantinaStoikou:
I fixed it for the other searches as well except the document search in the frontsite series details.

src/lib/components/backoffice/Sidebar/Sidebar.js
```diff
@@ -40,6 +40,8 @@ class Sidebar extends Component {
     const checkInActive = activePath.includes(BackOfficeRoutes.checkIn);
     const checkOutActive = activePath.includes(BackOfficeRoutes.checkOut);
 
+    const query = '&sort=mostrecent&p=1';
```
ntarocco, line 43: the issue is that it might be that we have some searches that do not support the `mostrecent` sorting. Do we really need this?

ntarocco (commented):
Thanks for fixing this! IMHO, we need to do some digging in RSK to double check how default values are handled :)

ntarocco:
I think the issue is in RSK, this one: https://github.com/inveniosoftware/react-searchkit/issues/110

src/lib/pages/frontsite/Home/Headline/Headline.js
```diff
@@ -51,7 +51,7 @@ const HeadlineLayout = props => {
                     className="headline-quick-access"
                     as={Link}
                     to={FrontSiteRoutes.documentsListWithQuery(
-                      '&sort=mostrecent&order=desc'
+                      '&sort=mostrecent&order=desc&p=1'
```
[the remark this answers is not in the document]
FlorianCassayre: I would probably go for that too, since the issue can still be reproduced if provided a url without the `p=1` parameter; ie. [https://localhost:3000/search?q=](https://localhost:3000/search?q=). Thanks Konstantina for working on it, it's a very annoying issue that has been here for a long time.

FlorianCassayre (commented):
For the sorting filters it's much more coherent now, good job. Here is some additional context to the issue (Karolina can correct me):
- By default a sorting filter is a pair of parameters: the field to compare (title, date) and a traversal order (asc, desc)
- For the frontsite we want this feature to me more user-friendly, therefore instead of having one input for each parameter we provide a unified one which carries both informations at the same time (ie. "Title [A-Z]"). It allows to have an explicit name for the sorting filter and we can also get rid of useless filters that do not belong to any reasonable use case (ie. "Best match + Desc" can be interpreted as "Worst match" which is not very useful...)
- For the backoffice we would like to allow more flexibility and independence, reason why there are two inputs. However it seems that some search pages have this input while others don't; I think we should stick to the same thing everywhere otherwise I find it confusing

Could you fix [this typo](https://github.com/inveniosoftware/react-invenio-app-ils/blob/master/src/lib/config/defaultConfig.js#L664) by the way? Thanks

src/lib/modules/SearchControls/SearchSortBy.js
```diff
@@ -15,20 +15,13 @@ export default class SearchSortBy extends Component {
...
         </Responsive>
         <Responsive {...Responsive.onlyMobile}>
           {searchConfig.SORT_BY.length > 0 ? (
-            <SortBy
-              values={searchConfig.SORT_BY}
-              defaultValue={searchConfig.SORT_BY[0].value}
-              defaultValueOnEmptyString={searchConfig.SORT_BY_ON_EMPTY_QUERY}
-              overridableId="mobile"
-            />
+            <SortBy values={searchConfig.SORT_BY} overridableId="mobile" />
```
ntarocco, line 24: I think that the SortBy/SortOrder should completely disappear, in favor of `Sort` only as in the frontiste. WDYT?

ntarocco:
>     * For the backoffice we would like to allow more flexibility and independence, reason why there are two inputs. However it seems that some search pages have this input while others don't; I think we should stick to the same thing everywhere otherwise I find it confusing

Yes exactly: @kprzerwa started to move from `SortBy`+ `SortOrder` to `Sort` component only (which combines the 2), but it was too much to apply it everywhere. So she started with frontsite.
Now, @KonstantinaStoikou should apply this to all backoffice too.

It is much easier, IMHO, to use on cmp as you have described then 2.

FlorianCassayre:
Todo: when RSK is release, update dependency

FlorianCassayre:
The major fixes were successfully re-tested with dependent PR #137. This PR is now waiting for `react-searchkit` to be released.

ntarocco (approved):

PR #216, opened 2020-10-01
forms: fix delete button overlap with dropdown

topless:
- upgrade Formik 2.1.6
- fix propTypes for Subform
- fix activeIndex for ObjectListField
- fix DeleteActionButton overlap with dropdown menus in forms

The delete button `x` of the url should be hidden behind the dropdown or the Role
Screenshot 2020-10-01 at 5 05 26 PM

### DIFF
--- a/src/semantic-ui/site/collections/form.overrides
+++ b/src/semantic-ui/site/collections/form.overrides
@@ -27,7 +27,6 @@ Form Overrides - REACT-INVENIO-APP-ILS
     position: absolute;
     right: -0.4em;
     top: -0.3em;
-    z-index: 9999;
 
     .button {
       box-shadow: none;
--- a/package-lock.json
+++ b/package-lock.json
@@ -7556,9 +7556,9 @@
       }
     },
     "formik": {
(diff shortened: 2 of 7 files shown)


jrcastro2 (approved):

ntarocco (approved):

src/semantic-ui/site/collections/form.overrides
```diff
@@ -27,7 +27,6 @@ Form Overrides - REACT-INVENIO-APP-ILS
     position: absolute;
     right: -0.4em;
     top: -0.3em;
-    z-index: 9999;
```
kprzerwa, line 30: did you check if something else didn't get broken because of removing it?

topless: I checked most of the Document form elements like `urls`, `internal_notes` and `keywords`. I have checked out also `locations` form and `Series`. After removing it it also plays well with the calendars apart from the dropdowns. I believe is a leftover from the first implementation.

kprzerwa (approved):

PR #220, opened 2020-10-05
Forms: improved submission and errors

### LINKED ISSUE #169: tooltip alignment in form validation
Probably not related to this issue, but I noticed the following:

image

The error tooltip of the form field component can mess up the alignment. Possible fix would be to move the tooltip to the bottom of the input (see `error.pointing` property).

_Originally posted by @FlorianCassayre in https://github.com/inveniosoftware/react-invenio-app-ils/pull/162#issuecomment-674033259_

### LINKED ISSUE #192: Backoffice: SeriesEditor mounted twice
Go to the backoffice, select a series. Click on "Edit series": each vocabulary field is loaded twice (two HTTP requests).
It turns out the `SeriesEditor` component does not follow the same pattern as the others and stores its state using redux. Check all editor components in case others are affected.

FlorianCassayre:
* disable all fields when form is in submitting state (the loading state on the form did not prevent its edition)
* forms did not have proper handling for network errors
* refactor form editors, fix mounting issues (closes #192)
* fix error label displayed above selector (closes #169)


FlorianCassayre:
Would like to include refactoring changes before review (and fix the problem where some forms are being mounted twice).

src/lib/authentication/pages/Login/LoginWithLocalAccount/LoginWithLocalAccount.js
```diff
@@ -91,11 +91,12 @@ class LoginWithLocalAccount extends Component {
       const errors = getIn(error, 'response.data.errors', []);
 
       if (_isEmpty(errors)) {
-        const message = getIn(error, 'response.data.message', null);
-        if (message) {
-          actions.setSubmitting(false);
-          actions.setErrors({ message });
+        let message = getIn(error, 'response.data.message', null);
```
jrcastro2, line 94: If we return 'Unknown error' as default in the [getIn](https://lodash.com/docs/4.17.15#get) we could avoid the `if`
`const message = getIn(error, 'response.data.message', 'Unknown error');`

src/lib/forms/core/BaseForm/BaseForm.js
```diff
@@ -41,13 +41,12 @@ export class BaseForm extends Component {
       const errors = getIn(error, 'response.data.errors', []);
 
       if (_isEmpty(errors)) {
-        const message = getIn(error, 'response.data.message', null);
-        if (message) {
-          actions.setSubmitting(false);
-          actions.setErrors({ message });
-        } else {
-          throw error;
+        let message = getIn(error, 'response.data.message', null);
```
jrcastro2, line 44: same as above:
`const message = getIn(error, 'response.data.message', 'Unknown error');`

src/lib/pages/backoffice/Document/DocumentForms/DocumentEditor/DocumentEditor.js
```diff
@@ -9,9 +9,10 @@ import { DocumentForm } from './DocumentForm';
 export class DocumentEditor extends Component {
   constructor(props) {
     super(props);
+
     this.state = {
       data: {},
-      isLoading: true,
+      isLoading: !!props.match.params.documentPid,
```
ntarocco, line 15: what is this solving? I can think only that there could be an issue if you render 2 forms in the same page...

ntarocco (commented):
Thank you for all the fixes!

src/lib/pages/backoffice/Document/DocumentForms/DocumentEditor/DocumentEditor.js
FlorianCassayre: It doesn't solve anything actually, I purposefully added it for clarity :slightly_smiling_face: (for the form creation component although the state is not used, it was always in 'loading' which was confusing)

src/lib/authentication/pages/Login/LoginWithLocalAccount/LoginWithLocalAccount.js
FlorianCassayre: If the response is defined then we should look at `error.response.data.message` (the message returned by the server).
Otherwise there was a network issue and we can obtain the message from `error.message` (this message is provided by the javascript client, usually "Network Error").
But you're right that I can nest them in a one-liner like so:
```javascript
const message = getIn(error, 'response.data.message', getIn(error, 'message', 'Unknown error'));
```

src/lib/pages/backoffice/Document/DocumentForms/DocumentEditor/DocumentEditor.js
kprzerwa: if it is not used, maybe it's better to remove it then?

src/lib/pages/backoffice/EItem/EItemForms/EItemEditor/EItemEditor.js
```diff
@@ -4,12 +4,23 @@ import { Error } from '@components/Error';
...
+      error: {},
+    };
+  }
+
   componentDidMount() {
-    const { fetchEItemDetails, match } = this.props;
+    const { match } = this.props;
     if (match.params.eitemPid) {
-      fetchEItemDetails(match.params.eitemPid);
+      this.fetchEItem(match.params.eitemPid);
```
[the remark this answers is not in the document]
FlorianCassayre: Ah right, I've used `withCancel` in the past but forgot to do it here. Thanks a lot!

src/lib/pages/backoffice/Document/DocumentForms/DocumentEditor/DocumentEditor.js
FlorianCassayre: @kprzerwa the state is used for edit forms but not for create since we do not need to fetch anything. It just feels wrong to have an incorrect value even though it is not used...

ntarocco (approved):

src/lib/pages/backoffice/Document/DocumentForms/DocumentEditor/DocumentEditor.js
kprzerwa: I see, so it is used in the end :) I understood from the previous comment that it isn't

kprzerwa (approved):

PR #221, opened 2020-10-05
errorHandling: Adds global error handler

jrcastro2:
* Adds global error handler for 404, 429 and 500
* Adds `errorId` prop (sentry id) to the error component

### 429 page
image

### DIFF
--- a/src/lib/api/base.js
+++ b/src/lib/api/base.js
@@ -1,6 +1,6 @@
 import { invenioConfig } from '@config';
 import { goTo } from '@history';
-import { AuthenticationRoutes } from '@routes/urls';
+import { AuthenticationRoutes, FrontSiteRoutes } from '@routes/urls';
 import axios from 'axios';
 
 const apiConfig = {
@@ -20,6 +20,8 @@ const urlShouldNotRedirect = url => {
 
 const http = axios.create(apiConfig);
 
+const errorsList = [404, 429, 500];
+
 const responseRejectInterceptor = error => {
   if (
     error.response &&
@@ -27,6 +29,17 @@ const responseRejectInterceptor = error => {
     !urlShouldNotRedirect(error.response.request.responseURL)
   ) {
     goTo(`${AuthenticationRoutes.login}?sessionExpired`);
+  }
+
+  if (error.response && errorsList.includes(error.response.status)) {
+    error.response.data.error_id
+      ? goTo(FrontSiteRoutes.errors, {
+          errorCode: error.response.status,
+          errorId: error.response.data.error_id,
+        })
+      : goTo(FrontSiteRoutes.errors, {
+          errorCode: error.response.status,
+        });
   }
 
   return Promise.reject(error);
--- a/src/lib/components/HttpErrors/HttpError.js
+++ b/src/lib/components/HttpErrors/HttpError.js
@@ -6,7 +6,7 @@ import Overridable from 'react-overridable';
 
 class HttpError extends Component {
   render() {
-    const { title, message, icon, isBackOffice } = this.props;
+    const { title, message, icon, isBackOffice, errorId } = this.props;
     return (
       <Overridable id="HttpError.layout" {...this.props}>
         <div className="frontsite">
@@ -20,6 +20,7 @@ class HttpError extends Component {
               <Icon name={icon} size="massive" />
               <h1>{title}</h1>
               <h3>{message}</h3>
+              {errorId ? <h4> Error Id: {errorId}</h4> : null}
               {!isBackOffice && (
                 <Link to="/">
                   <Button icon labelPosition="left" primary>
@@ -41,10 +42,12 @@ HttpError.propTypes = {
   message: PropTypes.string.isRequired,
   icon: PropTypes.string.isRequired,
   isBackOffice: PropTypes.bool,
+  errorId: PropTypes.string,
 };
 
 HttpError.defaultProps = {
   isBackOffice: false,
+  errorId: undefined,
 };
 
 export default Overridable.component('HttpError', HttpError);
--- a/src/lib/pages/frontsite/ErrorsPage/ErrorsPage.js
+++ b/src/lib/pages/frontsite/ErrorsPage/ErrorsPage.js
@@ -1,0 +1,26 @@
+import React, { Component } from 'react';
+import {
+  InternalServerError,
+  NotFound,
+  TooManyRequests,
+} from '@components/HttpErrors';
+
+export class ErrorsPage extends Component {
+  render() {
+    const params = window.history.state ? window.history.state.state : null;
+
+    if (params) {
+      if (params.errorCode >= 500) {
+        if (params.errorId) {
+          return <InternalServerError errorId={params.errorId} />;
+        }
+        return <InternalServerError />;
+      } else if (params.errorCode === 404) {
+        return <NotFound />;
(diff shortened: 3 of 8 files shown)


src/lib/api/base.js
```diff
@@ -29,6 +31,19 @@ const responseRejectInterceptor = error => {
     goTo(`${AuthenticationRoutes.login}?sessionExpired`);
   }
 
+  errorsList.forEach(errorCode => {
+    if (
+      error.response &&
+      error.response.status === errorCode &&
```
ntarocco, line 37: Instead of the forEach, why not:
```
error.response && [404, 429, 500].includes(error.response.status)
```
see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes

src/lib/authentication/utils.js
```diff
@@ -31,3 +31,7 @@ export const parseParams = (queryString = '') => {
   });
   return params;
 };
+
+export const parseParamsToString = (queryString = '') => {
+  return Qs.parse(queryString, { ignoreQueryPrefix: true });
```
ntarocco, line 36: Totally agree with creating a reusable utility, but then can you please replace the code where it is used with this new function? e.g. just above, line 25
I am not sure we should have this utils in `authentication`, I guess it has nothing to do with that... I think it makes sense to move to somewhere more "global"?

src/lib/components/HttpErrors/HttpError.js
```diff
@@ -20,6 +20,7 @@ class HttpError extends Component {
               <Icon name={icon} size="massive" />
               <h1>{title}</h1>
               <h3>{message}</h3>
+              <h4>{errorId ? 'Error Id: ' + errorId : null}</h4>
```
ntarocco, line 23: I would move the `h4` inside:
```suggestion
              {errorId ? `<h4>Error Id: ${errorId}</h4>` : null}
```

src/lib/api/base.js
```diff
@@ -29,6 +31,19 @@ const responseRejectInterceptor = error => {
...
+      !urlShouldNotRedirect(error.response.request.responseURL)
+    ) {
+      const link = error.response.data.error_id
+        ? `${FrontSiteRoutes.errors}?errorCode=${errorCode}&errorId=${error.response.data.error_id}`
+        : `${FrontSiteRoutes.errors}?errorCode=${errorCode}`;
+      goTo(link);
```
ntarocco, line 43: Instead of using the URL params, I would use state so it is hidden to the user:
```
goTo(link, { errorCode: ..., errorId: ...})
```

src/lib/pages/frontsite/ErrorsPage/ErrorsPage.js
```diff
@@ -0,0 +1,19 @@
+import React, { Component } from 'react';
+import { InternalServerError, NotFound } from '@components/HttpErrors';
+import { parseParamsToString } from '@authentication/utils';
+
+export class ErrorsPage extends Component {
+  render() {
+    const params = parseParamsToString(window.location.search);
+    if ('errorCode' in params) {
```
ntarocco, line 8: see my comment above to use history `state` instead of URL params. WDYT?

topless:
Great stuff! Maybe we can make the `errorId` a link to the actual sentry url. We can introduce a Sentry url variable in the config, if it doesn't exist already.

src/lib/api/base.js
```diff
@@ -29,6 +31,19 @@ const responseRejectInterceptor = error => {
...
   }
 
+  errorsList.forEach(errorCode => {
+    if (
+      error.response &&
+      error.response.status === errorCode &&
+      !urlShouldNotRedirect(error.response.request.responseURL)
+    ) {
+      const link = error.response.data.error_id
+        ? `${FrontSiteRoutes.errors}?errorCode=${errorCode}&errorId=${error.response.data.error_id}`
```
topless, line 41: why call it `errorId` since it is the `sentryId`?

src/lib/authentication/utils.js
jrcastro2: I deleted it since if we use the state instead of the URL params we don't need it anymore

jrcastro2:
> Great stuff! Maybe we can make the `errorId` a link to the actual sentry url. We can introduce a Sentry url variable in the config, if it doesn't exist already.

Thanks! IMHO adding a link to sentry is not really interesting since this error page usually will be displayed to the users and they don't have access to sentry. It's just a reference for us, for example, if they send us a screenshot, we get the id.

src/lib/api/base.js
jrcastro2: I asked myself the same question, but since we call it `errorId` in the [backend](https://github.com/inveniosoftware/invenio-app-ils/blob/master/invenio_app_ils/errors.py#L44) I just thought we should stick to it and IMHO i think it's descriptive enough (maybe I am wrong)

kprzerwa:
> > Great stuff! Maybe we can make the `errorId` a link to the actual sentry url. We can introduce a Sentry url variable in the config, if it doesn't exist already.
> 
> Thanks! IMHO adding a link to sentry is not really interesting since this error page usually will be displayed to the users and they don't have access to sentry. It's just a reference for us, for example, if they send us a screenshot, we get the id.

I agree, it will create a "dead" link to click on

src/lib/pages/frontsite/ErrorsPage/ErrorsPage.js
```diff
@@ -0,0 +1,23 @@
...
+  render() {
+    const params = window.history.state ? window.history.state.state : null;
+    if (window.history.state) {
+      console.log(window.history.state.state);
+    }
+
+    if (params) {
+      if (params.errorCode === 429 || params.errorCode === 500) {
+        if (params.errorId) {
+          return <InternalServerError errorId={params.errorId} />;
```
kprzerwa, line 14: I will be the naming police here again, but the name `<InternalServerError>` as a name refers only to 500, so I wouldn't name the component which handles more errors like this. It will be misleading.  I would have a separate component either to handle each code separately, or by range (4XX response codes - one component - client errors, 5XX another component - server errors).
In this case 429 is very specific, and very different than 404, so I would give them each one component,
for 5xx type of response code we could have f.e. `ServerError`

kprzerwa (commented):
Is the red rectangle a part of the page? If yes I would remove it from around the error code, since the info is only for us. It should not catch so much attention from the user.

jrcastro2:
> Is the red rectangle a part of the page? If yes I would remove it from around the error code, since the info is only for us. It should not catch so much attention from the user.

No, it's not part of the page. I added it so you can quickly recognise the important parts, sorry for causing the misunderstanding. I will update the image wihtout any editing

src/lib/pages/frontsite/ErrorsPage/ErrorsPage.js
```diff
@@ -0,0 +1,26 @@
...
+  NotFound,
+  TooManyRequests,
+} from '@components/HttpErrors';
+
+export class ErrorsPage extends Component {
+  render() {
+    const params = window.history.state ? window.history.state.state : null;
+
+    if (params) {
+      if (params.errorCode === 500) {
```
ntarocco, line 13:
```suggestion
      if (params.errorCode >= 500) {
```

ntarocco (approved):

kprzerwa (approved):

PR #232, opened 2020-10-09
changes and decrease ES delay

### LINKED ISSUE #144: Loan: Replace button not working properly
### Description
When replacing an item on loan with another physical copy, the page displays an error `TypeError: authors is undefined`. And after refreshing the replace was not correctly done, since the copy that was replaced is not displayed as an available copy anymore

ntarocco:
* adds missing search delay after loan replace item action
* renames delay function to searchReady to avoid collision with lodash
* decrease the search delay to 2 seconds
* closes #144


.env
```diff
@@ -1,6 +1,6 @@
 REACT_APP_UI_ILS_VERSION=$npm_package_version
 REACT_APP_INVENIO_VERSION=1.0.0a13
-REACT_APP_INVENIO_UI_URL=https://localhost:5000
-REACT_APP_INVENIO_REST_ENDPOINTS_BASE_URL=https://localhost:5000/api
+REACT_APP_INVENIO_UI_URL=https://127.0.0.1:5000
+REACT_APP_INVENIO_REST_ENDPOINTS_BASE_URL=https://127.0.0.1:5000/api
```
FlorianCassayre, line 4: Out of curiosity, why this change?

src/lib/config/defaultConfig.js
```diff
@@ -10,7 +10,7 @@ import {
 export const APP_CONFIG = {
   ENABLE_LOCAL_ACCOUNT_LOGIN: true,
   ENABLE_OAUTH_LOGIN: true,
-  ES_DELAY: 3000,
+  SEARCH_READY_DELAY: 2000,
```
kprzerwa, line 13: will it have any impact for "slower" machines?

kprzerwa (approved):

src/semantic-ui/site/modules/dropdown.overrides
```diff
@@ -28,7 +28,6 @@ Dropdown Overrides - REACT-INVENIO-APP-ILS
 
   .mobile-filters-dropdown.ui.dropdown.scrolling {
     .menu.transition.visible {
-      width: 70vw !important;
```
FlorianCassayre, line 31: I see. Was there overflow issues?

.env
ntarocco: it was creating issues with the CSRF token for me, backend running on 127.0.0.1 and react app on localhost (or the opposite)

src/lib/config/defaultConfig.js
ntarocco: it should not: by default ES has refresh time 1 sec, it should be independent of the machines...

src/semantic-ui/site/modules/dropdown.overrides
ntarocco: Yes, see:

FlorianCassayre: Ah indeed on large mobile devices, good job!

PR #234, opened 2020-10-12
Backoffice: Adds warning when creating a loan [invenio-app-ils#952]

### LINKED ISSUE #152: Display a warning when creating a new loan on overbooked document
**Scenario:** a document has 3 pending loan requests.
The librarian, by mistake, goes to the 3rd pending loan request, the last one created, and tries to checkout the physical copy.

**Expected:** a warning should be displayed saying that this is not the first request created (not respecting the order of requests for this document).

jrcastro2:
* depends on https://github.com/inveniosoftware/invenio-app-ils/pull/952
* closes #152

image


src/lib/modules/ESSelector/ESSelectorModal.js
```diff
@@ -60,6 +62,9 @@ export default class ESSelectorModal extends Component {
             onSelectionsUpdate={this.onSelectionsUpdate}
             {...this.props}
           />
+          {warningMessage ? (
```
kprzerwa, line 65: I think the message does not belong with a selector component. It should be above it maybe?  In this way you add complexity (in already complicated place ;)) to the reusable component, just to be used this one time.

kprzerwa (commented):
There are more places from which we create a loan (the loan details page itself, from the patron details page, and from the Actions-> Check-out on the sidebar

src/lib/pages/backoffice/Loan/LoanDetails/AvailableItems/AvailableItems.js
```diff
@@ -66,23 +72,54 @@ export default class AvailableItems extends Component {
...
+  handleCloseModal = () => {
+    this.setState({ modalOpen: false });
+  };
+
   checkoutItemButton(item, loan) {
-    const { performCheckoutAction, isActionLoading } = this.props;
+    const { documentPendingLoans, isActionLoading } = this.props;
+    documentPendingLoans.hits.sort((a, b) => {
+      return (
+        new Date(a.metadata.request_start_date) -
```
jrcastro2, line 100: I am not sure if we should use the `metadata.request_start_date` instead the `created` for sorting the array. I did it like this because it could be that someone created a request that will start a month later, given this situation I understand that we should order by the `request_start_date`

ntarocco: I think you should let Elasticsearch sort, so in the search query, add the `sort` param and the right `field` and `order`.

src/lib/pages/backoffice/Loan/LoanDetails/AvailableItems/AvailableItems.js
```diff
@@ -21,12 +21,18 @@ export default class AvailableItems extends Component {
...
+      modalOpen: false,
+      checkoutItem: null,
+    };
   }
 
   componentDidMount() {
-    const { fetchAvailableItems, loan } = this.props;
+    const { fetchAvailableItems, fetchPendingLoans, loan } = this.props;
     fetchAvailableItems(loan.metadata.document_pid);
+    fetchPendingLoans(loan.metadata.document_pid);
```
ntarocco, line 35: It is safer to chain the requests to avoid issues a multiple re-renderings. I think that to achieve this, you will have to modify the `fetchAvailableItems` (if used only here!!!), e.g. `fetchAvailableItemsWithPendingLoans`, and inside the action.js:
```
await itemApi.list(
        itemApi
          .query()
          .withDocPid(documentPid)
          .withStatus(invenioConfig.ITEMS.canCirculateStatuses)
          .availableForCheckout()
          .qs()
      );

await loans...

update Redux with the info you need, for example, for each item:

item["is_first_requested"] = True/False
```
and then use that `is_first_requested` to display the modal or not.

src/lib/pages/backoffice/Loan/LoanDetails/AvailableItems/AvailableItems.js
```diff
@@ -221,6 +259,31 @@ export default class AvailableItems extends Component {
         >
           <Loader isLoading={isLoading}>
             <Error error={error}>{this.renderTable()}</Error>
+            <Modal open={modalOpen}>
+              <Modal.Header>Overbooked literature!</Modal.Header>
+              <Modal.Content>
+                The{' '}
```
ntarocco, line 265: I would change the message, because if I am the librarian:
* I have the book in my hands, I know which book it is
* I want to know which one is the first loan request that I should fulfil
* It might actually not be overbooked, the GitHub issue was wrong!

The link should open **in a new tab**:
```
<Modal.Header>Antecedent loan request found!</Modal.Header>
<Modal.Content>There is an antecedent <link>pending loan request</link> for this literature. Are you sure that you want to checkout the copy for this request?</Modal.Content>
```

I don't like the `antecedent` word, but I cannot find anything else. We should say that there is a "previous" loan request created...

src/lib/pages/backoffice/Loan/LoanDetails/AvailableItems/AvailableItems.js
jrcastro2: I understand your point, just to clarify this, we need to modify the `fetchLoanDetails` instead the `fetchAvailableItems`, since we are searching for the first requested **loan**. Thanks !

FlorianCassayre:
Would it make sense to also provide `request_start_date` as a frontend search filter (since we now have it in the backend)? Like "Least recent requests" to process the requests in FIFO order.

ntarocco (approved):

jrcastro2:
> Would it make sense to also provide `request_start_date` as a frontend search filter (since we now have it in the backend)? Like "Least recent requests" to process the requests in FIFO order.

I think the librarians could make good use of it, so IMHO it makes sense to add it. Thanks!

ntarocco (approved):

PR #239, opened 2020-10-15
auth: provide config option for logout method

kprzerwa:


src/lib/authentication/services/AuthenticationService.js
```diff
@@ -20,7 +20,12 @@ class AuthenticationService {
 
   logout = () => {
     const logoutUrl = `${invenioConfig.APP.REST_ENDOINTS_BASE_URL}/logout`;
-    return http.post(logoutUrl);
+    if (invenioConfig.APP.LOGOUT_METHOD === 'GET') {
+      window.location = logoutUrl;
```
topless, line 24: I wouldn't assign a string to `window.location` which is an object. We can use the `replace` method.
```suggestion
      window.location.replace(logoutUrl);
```

src/lib/config/defaultConfig.js
```diff
@@ -25,6 +25,7 @@ export const APP_CONFIG = {
       className: '',
     },
   },
+  LOGOUT_METHOD: 'GET',
```
topless, line 28: It is not clear to me, when should I use `GET` and when `POST` for logout in my app.

ntarocco: I kind of agree that this is tricky, in the sense that it will actually depend on how the user logged in, and not a global option. I have added a comment about this in `invenio-oauthclient`, we should figure it out there first.

ntarocco (approved):
Needs `invenio-accounts` release + `invenio-app-ils` deps bump up

PR #240, opened 2020-10-15
test: UI testing with cypress

FlorianCassayre:


ntarocco (approved):
Wow! Thanks a lot! This is a huge step forward in making this app more robust, and I trust this much more than unit testing!

I think we can improve code organization and add some `id` around, but it is a fantastic starting point!

cypress/integration/frontsite/details.test.spec.js
```diff
@@ -0,0 +1,104 @@
...
+          is_requested: isRequested,
+        } = xhr.response.body;
+        expect(hasActiveLoan).to.be.false;
+        if (isRequested) {
+          cy.contains('Have it delivered to my office').click();
+
+          cy.get('input[name=selectedDate]').click();
+
+          cy.wait('@location').wait(1000); // Cannot avoid this delay
+          cy.get('td:not(.disabled) > span.suicr-content-item')
```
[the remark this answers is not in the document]
FlorianCassayre: I think it has to do with how `semantic-ui-react-calendar` is implemented because I had some inconsistencies when removing the delay (to be investigated further of course). Also a few lines above this, someone had a similar issue with another calendar library ([here](https://www.cypress.io/blog/2019/01/22/when-can-the-test-click/)), the click is registered, the calendar is dismissed but not date is selected; had to use a workaround.

cypress/integration/frontsite/details.test.spec.js
```diff
@@ -0,0 +1,104 @@
...
+  const goToSearchDocument = () => {
+    // Books + available for loan
+    cy.visit(
+      '/search?f=doctype%3ABOOK&f=availability%3Aavailable%20for%20loan'
+    );
+  };
+
+  it('can view document details and request loan if available', () => {
+    cy.server();
+    cy.route('GET', '/api/locations/*').as('location');
```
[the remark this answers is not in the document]
FlorianCassayre: `L12`: tell the framework to listen for `xhr` requests
`L13`: name this specific route pattern as `location`, which can later be used by calling `@location`. It's not possible to register the route on the fly, because of the way the framework works. Perhaps we can have a global store for all the patterns we are interested in listening to.

cypress/integration/backoffice/overview.test.spec.js
```diff
@@ -0,0 +1,30 @@
...
+    cy.contains('button.primary', 'Sign in').click();
+  };
+
+  it('should be linked from/to the frontsite, can logout', () => {
+    login();
+
+    cy.contains('Hector Nabu').click();
+    cy.contains('Backoffice').click();
+
+    cy.url().should('eq', Cypress.config().baseUrl + '/backoffice');
```
topless, line 15: nit: we can cache `const baseUrl = Cypress.config().baseUrl`
nit: which one do you find more readable?
```suggestion
    cy.url().should('eq', `${Cypress.config().baseUrl}/backoffice`);
```

FlorianCassayre:
Thanks for the feedback, I also have some potentially valuable feedback for you following my short experience with the framework. I'll recap my observations by tonight.

cypress/integration/backoffice/overview.test.spec.js
```diff
@@ -0,0 +1,30 @@
...
+
+    cy.contains('Go to Frontsite').click();
+
+    cy.url().should('eq', Cypress.config().baseUrl + '/');
+
+    cy.go('back');
+
+    cy.url().should('eq', Cypress.config().baseUrl + '/backoffice');
+
+    cy.contains('Sign out').click();
```
topless, line 26: disclaimer: I haven't use Cypress before

Proposal for idea, I see we are using the text of the html elements to refer to them. Would it make sense to create a shared structure like 
```
{
  pages: {
    login: {
      button: 'Sign In'
    }
  }
}
```
we will achieve reusability in our cases but most importantly if we change an element we will have to change it in one place only. Just a proposal.

topless (approved):
Great work 🥳 🎉 !!

cypress/integration/backoffice/overview.test.spec.js
FlorianCassayre: 100% agree! I dislike the `Cypress.config().baseUrl` boilerplate thing, there has to be a better way of doing it.

cypress/integration/backoffice/overview.test.spec.js
FlorianCassayre: That also a question I asked to myself, how to factor common logic in this framework. For instance a lot of the tests will require the user to be logged in; instead of signing in the user each time using the UI we could take a shortcut and perform a request to the backend instead.
I also saw a mention of hooks that are run before/after a test, this should be looked into details as it will help us share common actions / aliases.

FlorianCassayre:
Here are some tips that I can give:

- verify existence rather than absence. Keep in mind that the framework will proceed to the next command as soon as the predicate is verified. For commands like "text X is not equal to Y" or "element Z doesn't exist", the execution will actually be halted for a fixed amount of time (5s by default I think), which is not great. Moreover other issues can arise when using this pattern, so try to avoid it
- each "it" should work independently and must not make strong assumptions on the current state of the database. They are however allowed to have side-effects on the database
- avoid hardcoded delays: you can usually substitute them by a predicate that the framework should loop over and wait for (eg. an element becomes visible, a loader finishes, a request completes)
- some scenarios (such as the workflows) require test data before execution. I couldn't come up with a good solution to this problem, but I believe what works best for us is to simply create the data from the UI using cypress
- when possible try to select elements in a human-friendly manner. A good tip is to remember that someone else should be able to read the code and understand how the interface behaves - without looking at the UI. This will make the tests more robust to layout modifications. If a text changes, no problem we can easily fix it (plus, we don't change the texts very often)

The documentation is well designed, I recommend skimming it.

PR #245, opened 2020-10-19
global: change sort config

ntarocco:
For reviewers: there are some sort by that ideally would require to add both ASC and DESC (e.g. loan end date), but we need also to keep the text relatively short. What about using icons?

Ideas with emojis (it does not looks super nice...):
sort_by_emojys


kprzerwa:
> For reviewers: there are some sort by that ideally would require to add both ASC and DESC (e.g. loan end date), but we need also to keep the text relatively short. What about using icons?
> 
> Ideas with emojis (it does not looks super nice...):
> sort_by_emojys

I like the icons, emojis look a bit less "serious"

kprzerwa (approved):

PR #246, opened 2020-10-20
axios: retry request on CSRF missing cookie/token error

### LINKED ISSUE #230: Fix Sentry exceptions
- [x] Issue with acq: [see Sentry issue](https://cds-sentry.web.cern.ch/sentry/cds-ils/issues/196)
- [x] Issue with CSRF and stats: [see Sentry issue](https://cds-sentry.web.cern.ch/sentry/cds-ils/issues/158)
    The issue here is that the stats endpoint, a POST request, requires a CSRF token. Since this request is started as the same time of the others, the cookie with the CSRF token is not yet retrieved. The solution is to execute ANY POST request when the cookie has been retrieved, in particular after the `/api/me` that happens each time the app is loaded.
   * first fix is to make sure that we always have a valid CSRF:
     * option 1, preferable: retry request with new cookie, see [axios retry](https://github.com/axios/axios/issues/934)
     * option 2: refresh before expiration, see [here](https://excalidraw.com/#room=62d3e7f5aecb20acc585,_LRkB6r3mTqTXqqgp_70dw)
  * make the POST /api/stats request to happen after the others. The React cmp is mounted before the request to fetch the document is completed. This cmp should be mounted (and HTTP request) after. How to reproduce:
    * open a new **private** window
    * open the network tab
    * go to a document details page, e.g. https://<host>/literature/3v0mx-k0w30

ntarocco:
* closes #230

### DIFF
--- a/src/lib/api/base.js
+++ b/src/lib/api/base.js
@@ -1,5 +1,5 @@
 import { invenioConfig } from '@config';
-import { goTo } from '@history';
+import { goTo, replaceTo } from '@history';
 import { AuthenticationRoutes, FrontSiteRoutes } from '@routes/urls';
 import axios from 'axios';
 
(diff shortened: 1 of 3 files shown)


src/lib/api/base.js
```diff
@@ -9,37 +9,70 @@ const apiConfig = {
...
+    const isUnauthorized = errorStatus === 401;
+    const originalRequest = error.config;
+    const requestURL = originalRequest.url;
+
+    if (isUnauthorized && urlShouldRedirect(requestURL)) {
+      goTo(`${AuthenticationRoutes.login}?sessionExpired`);
+    } else {
+      const alreadyRetried = originalRequest._retry;
+      if (isCSRFError(errorStatus, errorResponse) && !alreadyRetried) {
+        originalRequest._retry = true;
```
kprzerwa, line 65: nice idea

kprzerwa (approved):

zzacharo (approved):
Nice job!!!

src/lib/api/base.js
```diff
@@ -9,37 +9,70 @@ const apiConfig = {
   xsrfCookieName: 'csrftoken',
   xsrfHeaderName: 'HTTP_X_CSRFTOKEN',
 };
+const http = axios.create(apiConfig);
 
+const HTTP_STATUS_CODES_WITH_ERROR_PAGE = [404, 429, 500];
 const URLS_NOT_TO_REDIRECT_IF_UNAUTHORIZED = ['/me', '/me/loans'];
+const CSRF_ERROR_REASON_NO_COOKIE = 'CSRF cookie';
+const CSRF_ERROR_REASON_BAD_TOKEN = 'CSRF token';
```
jrcastro2, line 17: LGTM! Just one minor question,does it makes sense to add the [error](https://github.com/inveniosoftware/invenio-rest/blob/master/invenio_rest/errors.py#L114) caused by the [expiration of the token](https://github.com/inveniosoftware/invenio-rest/blob/master/invenio_rest/csrf.py#L81) ? 
By adding something like `const CSRF_ERROR_REASON_TOKEN_EXPIRED = 'CSRF error';` we would retry the request instead of showing the `CSRF error` (I guess the users will not understand it) and it would be "transparent" for the user. WDYT?

jrcastro2 (approved):

src/lib/api/base.js
ntarocco: Isn't that happening if you are tampering the token? Then you are probably doing something malicious... do you see any other use case?

jrcastro2: If you leave it open for 1 hour and you come back the csrf token is expired, or if you only do get requests(for 1 hour) it will never get reloaded, it's not a common situation but I guess it could happen.

PR #254, opened 2020-10-27
Cypress: add more frontsite tests

KonstantinaStoikou:
Added tests for frontsite. 

About this [comment](https://github.com/inveniosoftware/react-invenio-app-ils/pull/240#discussion_r506142014) I thought wrapping the selectors in variables is easier e.g. `const header = '.ui.header'` instead of using ids. 

Removed some status checks e.g. `cy.wait('@loan').should(xhr => expect(xhr.status).to.eq(401));` since we should be testing the ui and not the backend. 

Added custom command to login by making a backend request instead of using the ui ([comment](https://github.com/inveniosoftware/react-invenio-app-ils/pull/240#discussion_r506170328)).

### DIFF
--- a/cypress/integration/frontsite/literatureDetails.test.spec.js
+++ b/cypress/integration/frontsite/literatureDetails.test.spec.js
@@ -1,12 +1,29 @@
 //https://127.0.0.1:3000/search?f=doctype%3ABOOK
 
 describe('frontsite literature details', () => {
-  const goToSearchDocument = () => {
-    // Books + available for loan
+  const goToAvailableBookDetails = () => {
     cy.visit(
       '/search?f=doctype%3ABOOK&f=availability%3Aavailable%20for%20loan'
     );
+    cy.get('.fs-book-card')
+      .first()
+      .click();
   };
+
+  const goToEBookDetails = () => {
+    cy.visit('/search?f=medium%3AELECTRONIC_VERSION');
+    cy.get('.fs-book-card')
+      .first()
+      .click();
+  };
+
+  it('should require login to request loan and access files', () => {
+    goToAvailableBookDetails();
+    cy.contains('Sign in to loan');
+    goToEBookDetails();
+    cy.contains('Sign in to loan');
+    cy.contains('Please login to see restricted files.');
+  });
 
   it('can view document details and request loan if available', () => {
     cy.server();
@@ -14,28 +31,17 @@ describe('frontsite literature details', () => {
     cy.route('GET', '/api/me/loans?document_pid=*').as('loan');
     cy.route('GET', '/api/document-requests/*').as('requests');
 
-    goToSearchDocument();
+    cy.login({ email: 'patron1@test.ch', password: '123456' });
 
-    cy.get('.fs-book-card')
-      .first()
-      .click();
-
-    cy.url().should('contain', Cypress.config().baseUrl + '/literature/');
-
-    cy.wait('@loan').should(xhr => expect(xhr.status).to.eq(401));
-
-    cy.contains('Sign in to loan').click();
-
-    cy.get('input#email').type('patron1@test.ch');
-    cy.get('input#password').type('123456');
-    cy.contains('button.primary', 'Sign in').click();
-
+    goToAvailableBookDetails();
     cy.url().should('contain', Cypress.config().baseUrl + '/literature/');
 
     cy.contains('Yannic Vilma').click();
     cy.contains('Your loans').click();
     cy.wait('@requests');
-    cy.get(':nth-child(2).statistic > .value').then($stat1 => {
+
+    const loanRequestCountSelector = ':nth-child(2).statistic > .value';
+    cy.get(loanRequestCountSelector).then($stat1 => {
       let expectedNextCount = parseInt($stat1.text());
 
       cy.go('back');
@@ -46,21 +52,24 @@ describe('frontsite literature details', () => {
           is_requested: isRequested,
         } = xhr.response.body;
         expect(hasActiveLoan).to.be.false;
-        if (isRequested) {
+
+        if (!isRequested) {
           cy.contains('Have it delivered to my office').click();
 
           cy.get('input[name=selectedDate]').click();
 
           cy.wait('@location').wait(1000); // Cannot avoid this delay
-          cy.get('td:not(.disabled) > span.suicr-content-item')
+          const dayFromDatePicker =
+            'td:not(.disabled) > span.suicr-content-item';
+          cy.get(dayFromDatePicker)
             .first()
             .type('{enter}');
 
           cy.contains('button.primary', 'Request').click();
-          cy.contains('div.positive.message', 'Request created');
 
-          cy.wait(1000); // ES delay
+          cy.wait(5000); // ES delay
 
+          cy.contains('.header', 'Request created');
           cy.reload();
 
           expectedNextCount++;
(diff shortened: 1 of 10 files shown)


cypress/integration/frontsite/literatureDetails.test.spec.js
```diff
@@ -1,41 +1,51 @@
 //https://127.0.0.1:3000/search?f=doctype%3ABOOK
 
 describe('frontsite literature details', () => {
-  const goToSearchDocument = () => {
-    // Books + available for loan
+  Cypress.Cookies.defaults({
+    preserve: 'csrftoken',
```
FlorianCassayre, line 5: This is to keep the CSRF token after visiting the login page once in order to avoid visiting it each time; am I right?

cypress/support/commands.js
```diff
@@ -0,0 +1,31 @@
+Cypress.Commands.add('login', user => {
```
FlorianCassayre, line 1: Excellent! :+1:
If we happen to use it a lot, we can create auxiliary methods `loginAsPatron`, `loginAsLibrarian`, etc.

cypress/integration/frontsite/literatureRequest.test.spac.js
```diff
@@ -0,0 +1,30 @@
+describe('literature request form', () => {
+  Cypress.Cookies.defaults({
+    preserve: 'csrftoken',
+  });
+
+  it('should redirect to login if not logged in', () => {
+    cy.shouldRedirectToLogin('/request');
+  });
+
+  it('should display form whe logged in', () => {
```
FlorianCassayre, line 10: typo

cypress/integration/frontsite/literatureRequest.test.spac.js
```diff
@@ -0,0 +1,30 @@
...
+    cy.url().should('eq', Cypress.config().baseUrl + '/request');
+    cy.contains('h1', 'Request new literature');
+  });
+
+  it('should submit form', () => {
+    cy.login({ email: 'patron1@test.ch', password: '123456' });
+    cy.visit('/request');
+
+    cy.get('input#title').type('A new book');
+    cy.get('input#medium').click({ force: true });
```
FlorianCassayre, line 22: Doesn't it work without `force`?
I think it's discouraged to use this parameter (if possible) because it bypasses the visibility/obstruction checks. For instance the <kbd>X</kbd> button in form arrays is currently unclickable because 'hidden' by another element; and we really want to catch these kinds of bugs.

FlorianCassayre (commented):
I think you did a great job there, thanks for revising the current specs. Your remarks are very valid, I'm noting them down.

cypress/support/commands.js
KonstantinaStoikou: Yes, I agree, also it would be nice if we can hardcode some scenarios for some patrons, e.g. when we are testing the extension button we need to use a patron that will have a loan that can be extended.

cypress/integration/frontsite/literatureDetails.test.spec.js
KonstantinaStoikou: Yes, it's because cypress cleans all the cookies before each test, and the login command has to use the csrf cookie in the request headers.

cypress/integration/frontsite/literatureRequest.test.spec.js
```diff
@@ -0,0 +1,34 @@
+describe('literature request form', () => {
+  Cypress.Cookies.defaults({
```
topless, line 2: Instead of redefining this, Cypress has support for [global-default-cookies](https://docs.cypress.io/api/cypress-api/cookies.html#Set-global-default-cookies)

cypress/integration/frontsite/search.test.spec.js
```diff
@@ -23,21 +23,29 @@ describe('frontsite search', () => {
...
+    cy.url().should('contain', Cypress.config().baseUrl + '/literature');
+  });
 
   it('should handle empty results', () => {
     goToSearch();
     cy.get('@input').type('zzzzzz');
     cy.get('@search').click();
     cy.contains('No results found');
-    cy.contains('Availability').should('not.exist'); // See https://github.com/inveniosoftware/react-invenio-app-ils/issues/5
+    // cy.contains('Availability').should('not.exist'); // See https://github.com/inveniosoftware/react-invenio-app-ils/issues/5
```
topless, line 40: It is not very clear to me what should happen here? Should it be enabled when the issue is resolved? What is the expected behaviour?

KonstantinaStoikou: Right now if this line is uncommented the test will fail because of the issue (the availability filter is displayed even when the results are 0). When it is resolved we can uncomment it, or remove completely this check, or do this check for all filters.

FlorianCassayre: This was just to highlight the fact that there is bug, and such that once it is solved the line can be uncommented and the test should pass.

Capture d’écran de 2020-10-27 17-39-15

cypress/integration/frontsite/literatureDetails.test.spec.js
```diff
@@ -46,21 +52,24 @@ describe('frontsite literature details', () => {
...
+            'td:not(.disabled) > span.suicr-content-item';
+          cy.get(dayFromDatePicker)
             .first()
             .type('{enter}');
 
           cy.contains('button.primary', 'Request').click();
-          cy.contains('div.positive.message', 'Request created');
 
-          cy.wait(1000); // ES delay
+          cy.wait(5000); // ES delay
```
ntarocco, line 70:
```suggestion
          cy.wait(2000); // ES delay
```
It is true that `1` sec might not be enough, but `2` should be ok :)

cypress/integration/frontsite/profile.test.spec.js
```diff
@@ -0,0 +1,79 @@
...
+    cy.login({ email: 'patron1@test.ch', password: '123456' });
+    cy.visit('/profile');
+    cy.url().should('eq', Cypress.config().baseUrl + '/profile');
+
+    if (cy.get('table').contains('Cancel request')) {
+      cy.get('table').within(() => {
+        cy.contains('.button', 'Cancel request').click();
+      });
+      cy.contains('Are you sure you want to cancel your request?');
+      cy.contains('Yes, I am sure').click();
```
ntarocco, line 54: Minor: we might want to assert that it succeeded?

cypress/integration/frontsite/profile.test.spec.js
```diff
@@ -0,0 +1,79 @@
...
+    cy.url().should('eq', Cypress.config().baseUrl + '/profile');
+
+    const loanRequestDiv = '.ui.divided.items';
+
+    if (cy.get(loanRequestDiv).contains('Cancel request')) {
+      cy.get(loanRequestDiv).within(() => {
+        cy.contains('.button', 'Cancel request').click();
+      });
+      cy.contains('Are you sure you want to cancel your loan request?');
+      cy.contains('Yes, I am sure').click();
```
ntarocco, line 40: Minor: we might want to assert that it succeeded?

ntarocco (approved):
Impressive work! Thanks a lot @KonstantinaStoikou 💯 👍 👏 🚀

cypress/integration/frontsite/profile.test.spec.js
KonstantinaStoikou: I agree, the cancel buttons need further testing to check if the request has been removed from the profile page.

cypress/integration/frontsite/profile.test.spec.js
KonstantinaStoikou: Same as above.

PR #256, opened 2020-10-27
pagination: hide last page item

FlorianCassayre:
Tiny PR that aims to revise pagination. It does not make sense to display a button to go to the last page, firstly because in production the last page may not correspond to the last data entry but rather the last match within the allowed window (currently set at 10k).
Semantic-UI does not provide a way to avoid displaying a button for the last page, we can only control the side buttons; thus as suggested by @kprzerwa if we want to avoid rewriting the entire component the easiest solution is to fix it using CSS.

* First modification is to disable the <kbd><<</kbd> and <kbd>>></kbd> buttons, directly in the overriden component (which was conveniently tweaked to be the global pagination component, in a recent PR)
* Second one is to use a CSS selector that searches for a <kbd>…</kbd> at position 3 starting from the right hand side, and hides the next sibling (which turns out to be the last page)

Preview of the new look (for a result set that fits in 11 displayable pages):

Capture d’écran de 2020-10-27 13-46-16
Capture d’écran de 2020-10-27 13-46-03
Capture d’écran de 2020-10-27 13-45-41
Capture d’écran de 2020-10-27 14-04-05

### DIFF
--- a/src/lib/components/Pagination/Pagination.js
+++ b/src/lib/components/Pagination/Pagination.js
@@ -12,7 +12,6 @@ class Pagination extends Component {
       currentSize = invenioConfig.APP.DEFAULT_RESULTS_SIZE,
       onPageChange,
       loading,
-      simple,
       firstItem,
       lastItem,
       totalPages: _,
@@ -32,8 +31,8 @@ class Pagination extends Component {
           <Paginator
             activePage={currentPage}
             totalPages={totalDisplayedPages}
-            firstItem={simple ? null : firstItem}
-            lastItem={simple ? null : lastItem}
+            firstItem={null}
+            lastItem={null}
             onPageChange={(event, { activePage }) => onPageChange(activePage)}
             {...extraParams}
           />
@@ -50,13 +49,8 @@ Pagination.propTypes = {
   totalResults: PropTypes.number.isRequired,
   onPageChange: PropTypes.func.isRequired,
   loading: PropTypes.bool,
-  simple: PropTypes.bool,
   firstItem: PropTypes.any,
   lastItem: PropTypes.any,
-};
-
-Pagination.defaultProps = {
-  simple: false,
 };
 
 export default Overridable.component('Pagination', Pagination);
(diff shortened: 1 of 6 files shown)


src/lib/components/Pagination/Pagination.js
```diff
@@ -32,8 +32,8 @@ class Pagination extends Component {
           <Paginator
             activePage={currentPage}
             totalPages={totalDisplayedPages}
-            firstItem={simple ? null : firstItem}
-            lastItem={simple ? null : lastItem}
+            firstItem={null}
```
topless, line 35: With this change the `simple` prop is not supported any longer, we should remove it from the `PropTypes` and `defaultProps` and also from the components 

- `ExistingRelations.js`
- `frontsite/PatronProfile/LoanList.js`
- `PatronCurrentDocumentRequests.js`
- `PatronPastDocumentRequests.js`

FlorianCassayre: Yes, I'm the one who added it in the first place; I was aware of that and left it voluntarily in case we want to have specific logic for smaller screens. I'll remove it if no one is against.

src/semantic-ui/site/globals/site.overrides
```diff
@@ -137,6 +137,11 @@ a {
   justify-content: center;
 }
 
+.pagination [type=ellipsisItem]:nth-last-child(3) + [type=pageItem] {
```
ntarocco, line 140: To double check if it makes sense and test it:
```suggestion
.ui.pagination [type=ellipsisItem]:nth-last-child(3) + [type=pageItem] {
```
just because I saw that there are 2 paginations cmps and classes:
```
<div class="pagination">
    <div class="ui pagination menu" ...>
```

ntarocco (approved):

src/semantic-ui/site/globals/site.overrides
FlorianCassayre: Thanks, it's indeed safer to specify the `ui` class.

FlorianCassayre:
Ran tests locally.

PR #258, opened 2020-10-28
fix price field

### LINKED ISSUE #155: PriceField component - Fix validation
The `PriceField` component is not validating properly the data, and not cleaning the data when submitting an empty value, this can causes a bug and it doesn't display any information. This component is not setting default values correctly, in some places there should be a default currency assigned but it is not. This bug doesn't allow to submit the form. Example in `purchase order` form.
image

the bug is blocking feature:
 Create new order: it is not working if we add payment information due to `price.currency` is not set   inveniosoftware/react-invenio-app-ils#155 if we don't supply the price though `<OrderStatistics>` component   throws `Uncaught TypeError: Cannot read property 'currency' of undefined` in details

FlorianCassayre:
* closes #155

The following changes and simplifications were made to the price field in order to finally make it work correctly:
* the serializer recursively clears the nested objects with empty values
  * for instance `{ a: { b: '' } }` will become `{}`
  * this also fixes other issues with accordions that were touched but eventually left unfilled and collapsed
* in case of a fixed currency, the feature of displaying a label was discarded in favor of the dropdown
  * instead of displaying all currencies the dropdown only shows `defaultCurrency` 
* the currency dropdown always contains an empty value <kbd>-</kbd> and defaults to that
* semantic-UI has bogus error popups for grouped inputs; we now make sure the field is at least highlighted red on any error

As an illustration, the following price field is correctly filled:

These two are considered non-empty - and incorrect (respectively, the value field is nonempty and the currency dropdown is nonempty):

While this one is considered empty (and treated as such):

### DIFF
--- a/src/lib/forms/core/PriceField.js
+++ b/src/lib/forms/core/PriceField.js
@@ -1,13 +1,13 @@
 import { FastField, Field, getIn } from 'formik';
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
-import { Dropdown, Form, Input, Label } from 'semantic-ui-react';
+import { Dropdown, Form, Input } from 'semantic-ui-react';
 import { invenioConfig } from '@config';
 import { vocabularyApi } from '@api/vocabularies';
 import { withCancel } from '@api/utils';
 import _isEmpty from 'lodash/isEmpty';
 
-class PriceDropdown extends Component {
+class CurrencyDropdown extends Component {
   state = {
     isLoading: true,
     error: null,
@@ -53,23 +53,20 @@ class PriceDropdown extends Component {
   };
 
   getAllOptions = (options, value) => {
-    const {
-      field: { required },
-    } = this.props;
+    const { defaultCurrency } = this.props;
     const { isLoading } = this.state;
-    if (!required) {
-      options = [
-        {
-          key: '',
-          value: undefined,
-          text: '-',
-        },
-        ...options,
-      ];
-    }
+    const visibleOptions =
+      defaultCurrency !== null
+        ? options.filter(o => o.value === defaultCurrency)
+        : options;
+    visibleOptions.unshift({
+      key: '',
+      value: '',
+      text: '-',
+    });
     if (!isLoading) {
-      if (!_isEmpty(value) && !options.find(o => o.value === value)) {
-        options.push({
+      if (!_isEmpty(value) && !visibleOptions.find(o => o.value === value)) {
+        visibleOptions.push({
           key: value,
           value: value,
           text: `Missing value: ${value}`,
@@ -77,7 +74,7 @@ class PriceDropdown extends Component {
         });
       }
     }
-    return options;
+    return visibleOptions;
   };
 
   render() {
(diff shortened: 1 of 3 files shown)


src/lib/forms/core/PriceField.js
```diff
@@ -131,18 +130,21 @@ export class PriceField extends Component {
...
       canSelectCurrency,
       required,
     } = this.props;
     const name = `${fieldPath}.currency`;
-    return canSelectCurrency ? (
-      <Field name={name} required={required} component={PriceDropdown} />
-    ) : (
-      <Label name={name}>{defaultCurrency}</Label>
+    return (
+      <Field
```
topless, line 142: nit: If I understood correctly from your issue description when we have a fixed currency the dropdown will have only one value. I see that [dropdown](https://react.semantic-ui.com/modules/dropdown/) supports also `disabled` prop, we could add that in case we have a fixed currency.

src/lib/forms/core/PriceField.js
```diff
@@ -53,20 +53,20 @@ class PriceDropdown extends Component {
...
+        ? options.filter(o => o.value === defaultCurrency)
+        : options;
+    options = [
+      {
+        key: '',
+        value: '',
+        text: '-',
+      },
+      ...otherOptions,
+    ];
```
topless, line 69: Instead of redefining `options` which is our argument we can do something like
```suggestion
    otherOptions.unshift({key: '', value: '', text: '-' })
```

topless (approved):

src/lib/forms/core/PriceField.js
FlorianCassayre: The thing is, if the field is optional and the user wants to leave the field empty they must select the <kbd>-</kbd> option, otherwise this value will be sent to & refused by the server.

src/lib/forms/core/PriceField.js
FlorianCassayre: I'm also not a fan of reassigning a value to the argument.

src/lib/forms/core/PriceField.js
ntarocco: I think it is a compromise that we can accept, at least until we have anything better.
Just to understand:
* by default, when creating a new record (empty form), currency fields will be with `-`, so the user will have to choose something if required and/or price has a value.
* when editing, the currency field has a value. Only when clearing the price value (which is rare I guess), then the user will have to set `-` in the currency field.
Is this correct?

ntarocco (approved):

FlorianCassayre:
* yes, always <kbd>-</kbd> by default. If the user wants to fill the price they have to manually select the currency aswell, even if there is only one possible nonempty option. In a way this also helps prevent typos but I would not describe it as a feature
* yes, that's also correct

:+1: Of course I will be more than happy if someone comes up with a better UX. But we after spending decent amount of time on it without success we must focus on making it work first.

FlorianCassayre:
[travis build](https://travis-ci.com/github/FlorianCassayre/react-invenio-app-ils/builds/194815236) (currently running).

FlorianCassayre:
[travis build](https://travis-ci.com/github/FlorianCassayre/react-invenio-app-ils/builds/195364631) (passing).

PR #275, opened 2020-11-05
CheckoutSearch: get results for email without quotes

KonstantinaStoikou:
When searching by an email in the checkout page before there were no results at all when you didn't wrap the email in quotes. Now it shows results without quotes.
Closes partially CERNDocumentServer/cds-ils#214

### DIFF
--- a/src/lib/api/patrons/patron.js
+++ b/src/lib/api/patrons/patron.js
@@ -33,7 +33,7 @@ class QueryBuilder {
     this.patronQuery = [];
   }
 
-  withEmail(email, wildcard = false) {
+  withEmail(email) {
     if (!email) {
       throw TypeError('Email argument missing');
     }
--- a/src/lib/api/patrons/patron.js
+++ b/src/lib/api/patrons/patron.js
@@ -37,7 +37,7 @@ class QueryBuilder {
     if (!email) {
       throw TypeError('Email argument missing');
     }
-    this.patronQuery.push(`email:${email}${wildcard ? '*' : ''}`);
+    this.patronQuery.push(`email:${email}`);
     return this;
   }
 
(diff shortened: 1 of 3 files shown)


src/lib/api/patrons/patron.js
```diff
@@ -37,7 +37,7 @@ class QueryBuilder {
     if (!email) {
       throw TypeError('Email argument missing');
     }
-    this.patronQuery.push(`email:${email}${wildcard ? '*' : ''}`);
+    this.patronQuery.push(`email:${email}`);
```
ntarocco, line 40: I am not 100% sure about this change, the `*` was there for something probably? And this is changing any usage of this query builder, not only `CheckoutSearch`.

Probably the removal of the `*` is good (did you test many different cases?), but wouldn't be better in the checkout search to avoid to specify fields? We are limiting to email and name, but we need to support the person_id too...

ntarocco: Actually I have double check, and we are missing the config `PATRONS.customFields` in `cds-ils`, see [here](https://github.com/inveniosoftware/react-invenio-app-ils/blob/e1d10e64ef485c2ddcebd64c6ba4617040080f6e/src/lib/config/defaultConfig.js#L794) and how it is used [here](https://github.com/inveniosoftware/react-invenio-app-ils/blob/e1d10e64ef485c2ddcebd64c6ba4617040080f6e/src/lib/pages/backoffice/Actions/CheckOut/state/actions.js#L38).
Can you add a PR in `cds-ils` to add the config?

ntarocco, line 40: Please remove the argument `wildcard` then now

kprzerwa (approved):

ntarocco (approved):

PR #277, opened 2020-11-05
ILSMenu: add button to go to the search page

### LINKED ISSUE #259: Add 'Catalogue' to the top menu, pointing to the search page
During user testing, there was sometimes missing a way to navigate quickly back to the search page

KonstantinaStoikou:
Closes #259

Screenshot 2020-11-06 at 16 17 13
Mobile:
Screenshot 2020-11-06 at 16 30 21

### DIFF
--- a/src/lib/components/ILSMenu/ILSMenu.js
+++ b/src/lib/components/ILSMenu/ILSMenu.js
@@ -12,6 +12,7 @@ import {
   Dropdown,
   Image,
   Menu,
+  Icon,
   Responsive,
 } from 'semantic-ui-react';
 
@@ -115,6 +116,11 @@ class ILSMenu extends Component {
                   </div>
                 </Menu.Item>
                 <Menu.Menu position="right">
+                  <Menu.Item>
+                    <Link to={FrontSiteRoutes.documentsListWithQuery('')}>
+                      Search
+                    </Link>
+                  </Menu.Item>
                   <Overridable id="ILSMenu.RightMenuItems" />
                   <Menu.Item>
                     {this.renderRightMenuItem(user.fullName)}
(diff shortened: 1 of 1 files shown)


src/lib/components/ILSMenu/ILSMenu.js
```diff
@@ -115,6 +116,13 @@ class ILSMenu extends Component {
                   </div>
                 </Menu.Item>
                 <Menu.Menu position="right">
+                  <Menu.Item
+                    onClick={() =>
```
FlorianCassayre, line 120: In general we should use `Link` when creating links, for accessibility reasons, instead of relying on events. For instance here it's not possible to open it in a new tab, while I'm sure some users may want to do so.

src/lib/components/ILSMenu/ILSMenu.js
```diff
@@ -115,6 +116,13 @@ class ILSMenu extends Component {
...
+                      goTo(FrontSiteRoutes.documentsListWithQuery(''))
+                    }
+                  >
+                    Go to the catalogue
```
FlorianCassayre, line 124: What about simply `Catalogue`? (to reduce the complexity of the menu)

KonstantinaStoikou: In cds-ils the logo says `CERN Library Catalogue` and I thought having a second button that just says `Catalogue` would be confusing.

ntarocco: What about `Search`? I would be in favor of one word too... maybe quick chat with the team? during stand up?

FlorianCassayre:
One more thing that I thought of this morning: how does it render on mobile devices?

(on the latest `master` the environment label size was reduced on mobiles, so there may be some space left now. Also beware that the `cds-ils` logo is much wider than the standard ILS placeholder)

kprzerwa (approved):

FlorianCassayre:
Small remark: the link is harder to click than its neighbouring dropdown.

PR #284, opened 2020-11-06
backoffice: update location form opening hours

FlorianCassayre:
* closes inveniosoftware/invenio-app-ils#980, assuming inveniosoftware/invenio-app-ils#981 is already merged (but AFAIK no blocking dependency)

For the sake of simplicity, the two period fields are always there. The user may leave a pair of inputs empty, and it will be filtered out. The consequence is that if the period is defined in the second pair, it will be moved to the first (as illustrated below).

(before submission)

(after submission, when editing again)

### DIFF
--- a/src/lib/pages/backoffice/Location/LocationForms/LocationEditor/components/LocationForm/Closures/Closures.js
+++ b/src/lib/pages/backoffice/Location/LocationForms/LocationEditor/components/LocationForm/Closures/Closures.js
@@ -1,22 +1,33 @@
 import React, { Component } from 'react';
-import { Segment, Grid, Header, Message } from 'semantic-ui-react';
+import {
+  Segment,
+  Grid,
+  Header,
+  Message,
+  Button,
+  Icon,
+} from 'semantic-ui-react';
+import PropTypes from 'prop-types';
 import { BooleanField } from '@forms/core/BooleanField';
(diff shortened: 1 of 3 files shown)


jrcastro2 (approved):
LGTM! Minor: This fixes the problem but doesn't seems really intuitive as a user, maybe we could improve this in the future by just showing only one period and a button that will allow to add another period.

FlorianCassayre:
Yes I agree. Since there is now a form refactoring task, I will include it there.

kprzerwa (commented):
I agree with @jrcastro2,
IMO it should be fixed here not in the forms. It happens only in this form, in it not a part of global refactoring of forms.

ntarocco:
> I agree with @jrcastro2,
> IMO it should be fixed here not in the forms. It happens only in this form, in it not a part of global refactoring of forms.

I understand, but I am not sure it is worth it?
What about if we simple add a text, somewhere in the second part of the period of the day, explaining? e.g.
`if you need to add one period, just fill in the first or optionally insert both for 2 periods` (better English please, but you got the point!)

FlorianCassayre:
Here is how it looks now:

image

ntarocco (approved):

kprzerwa (approved):

PR #289, opened 2020-11-09
itemForm: Improves item edit

jrcastro2:
* Adds warning modal when updating the document
  of an item that has past loans
* Disables edit button when item on loan
* closes https://github.com/inveniosoftware/invenio-app-ils/issues/228

While editing an item if it has past loans the form will display the document field disabled and with a informative message.

image

### DIFF
--- a/src/lib/pages/backoffice/Item/ItemForms/ItemEditor/ItemForm/ItemForm.js
+++ b/src/lib/pages/backoffice/Item/ItemForms/ItemEditor/ItemForm/ItemForm.js
@@ -99,8 +99,17 @@ export class ItemForm extends Component {
       successSubmitMessage,
       title,
       pid,
+      itemPastLoans,
     } = this.props;
     const initialValues = data ? this.prepareData(metadata) : {};
+    const hasPastOrActiveLoans =
+      itemPastLoans > 0 ||
+      (data.metadata.circulation &&
+        invenioConfig.CIRCULATION.loanActiveStates.includes(
+          data.metadata.circulation.state
+        ));
+    const message =
+      'Change of document is disabled because this item has active or past loans.';
     return (
       <BaseForm
         initialValues={{
@@ -134,6 +143,8 @@ export class ItemForm extends Component {
               icon={<DocumentIcon />}
               query={documentApi.list}
               serializer={serializeDocument}
+              disabled={hasPastOrActiveLoans}
+              disabledMessage={hasPastOrActiveLoans ? message : ''}
               width={8}
             />
             <VocabularyField
@@ -246,6 +257,7 @@ ItemForm.propTypes = {
   successSubmitMessage: PropTypes.string,
   title: PropTypes.string,
   pid: PropTypes.string,
+  itemPastLoans: PropTypes.number,
 };
 
 ItemForm.defaultProps = {
@@ -253,4 +265,5 @@ ItemForm.defaultProps = {
   successSubmitMessage: null,
   title: null,
   pid: null,
+  itemPastLoans: null,
 };
--- a/src/lib/pages/backoffice/Item/ItemForms/ItemEditor/index.js
+++ b/src/lib/pages/backoffice/Item/ItemForms/ItemEditor/index.js
@@ -1,1 +1,18 @@
-export { ItemEditor } from './ItemEditor';
+import { connect } from 'react-redux';
+
+import { fetchPastLoans } from '../../ItemDetails/ItemPastLoans/state/actions';
+import ItemEditorComponent from './ItemEditor';
+
+const mapStateToProps = state => ({
+  itemPastLoans: state.itemPastLoans.data.total,
+  isLoadingPastLoans: state.itemPastLoans.isLoading,
+});
+
+const mapDispatchToProps = dispatch => ({
+  fetchPastLoans: itemPid => dispatch(fetchPastLoans(itemPid)),
+});
+
+export const ItemEditor = connect(
+  mapStateToProps,
+  mapDispatchToProps
+)(ItemEditorComponent);
--- a/src/lib/forms/core/BaseForm/BaseForm.js
+++ b/src/lib/forms/core/BaseForm/BaseForm.js
@@ -151,4 +151,15 @@ BaseForm.propTypes = {
 
 BaseForm.defaultProps = {
   children: null,
+  initialValues: null,
+  successSubmitMessage: '',
+  successCallback: null,
+  createApiMethod: null,
+  editApiMethod: null,
+  submitSerializer: null,
+  title: '',
+  pid: '',
+  validationSchema: null,
+  onSubmit: null,
+  buttons: null,
 };
--- a/src/lib/forms/core/SelectorField.js
+++ b/src/lib/forms/core/SelectorField.js
@@ -85,6 +85,7 @@ export class SelectorField extends Component {
       renderGroup,
       renderSelection,
       width,
+      disabledMessage,
       ...selectorProps
     } = this.props;
     const selections = [];
@@ -123,6 +124,7 @@ export class SelectorField extends Component {
           initialSelections={selections}
           renderSelections={renderGroup}
           renderSelection={renderSelection}
+          disabledMessage={disabledMessage}
           onSelectionsUpdate={selections =>
             this.onSelectionsUpdate(selections, setFieldValue)
           }
@@ -164,6 +166,7 @@ SelectorField.propTypes = {
   multiple: PropTypes.bool,
   placeholder: PropTypes.string,
   width: PropTypes.number,
+  disabledMessage: PropTypes.string,
 };
 
 SelectorField.defaultProps = {
@@ -178,4 +181,5 @@ SelectorField.defaultProps = {
   renderGroup: null,
   label: '',
   icon: null,
+  disabledMessage: '',
 };
--- a/src/lib/modules/ESSelector/ESSelector.js
+++ b/src/lib/modules/ESSelector/ESSelector.js
@@ -1,7 +1,14 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import Overridable from 'react-overridable';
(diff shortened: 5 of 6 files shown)


src/lib/pages/backoffice/Item/ItemDetails/ItemActionMenu/ItemActionMenu.js
```diff
@@ -113,16 +113,32 @@ export default class ItemActionMenu extends Component {
...
 
     return (
       <div className="bo-action-menu">
-        <EditButton
-          fluid
-          to={BackOfficeRoutes.itemEditFor(item.metadata.pid)}
-          text="Edit physical copy"
-        />
+        {isInCirculation ? (
+          <Popup
```
jrcastro2, line 122: Does it makes sense to create a `FakeButton` or `DisabledButton` component with this `PopUp` code and replace all the disables button with this, that way the user will have some feedback on why the button is disabled. WDYT ?

ntarocco: Is this done because the Popup is not displayed on disabled buttons?
I think the button should be disabled.

src/lib/pages/backoffice/Item/ItemDetails/ItemActionMenu/ItemActionMenu.js
```diff
@@ -164,9 +180,11 @@ ItemActionMenu.propTypes = {
   deleteItem: PropTypes.func.isRequired,
   checkoutItem: PropTypes.func.isRequired,
   checkOutitemList: PropTypes.arrayOf(PropTypes.object),
+  isInCirculation: PropTypes.bool,
 };
 
 ItemActionMenu.defaultProps = {
   offset: 0,
   checkOutitemList: [],
+  isInCirculation: false,
```
ntarocco, line 189: I would go maybe for something more explicit?
```suggestion
  isCurrentlyOnLoan: false,
```

src/lib/pages/backoffice/Item/ItemForms/ItemEditor/ItemForm/ItemForm.js
```diff
@@ -92,15 +92,36 @@ export class ItemForm extends Component {
     goTo(BackOfficeRoutes.itemDetailsFor(getIn(response, 'data.metadata.pid')));
   };
 
+  shouldModalDisplay(initialValues, values) {
```
ntarocco, line 95: I would go for:
```suggestion
  isDocumentPidChanged(initialValues, values) {
```

src/lib/pages/backoffice/Item/ItemForms/ItemEditor/ItemForm/ItemForm.js
```diff
@@ -92,15 +92,36 @@ export class ItemForm extends Component {
...
     } = this.props;
     const initialValues = data ? this.prepareData(metadata) : {};
+    const itemOnLoan =
+      data && data.metadata && data.metadata.circulation
+        ? invenioConfig.CIRCULATION.loanActiveStates.includes(
+            data.metadata.circulation.state
+          )
+        : false;
+    const modal = {
+      header: 'Possible inconsistency',
```
ntarocco, line 120: Just wondering: what about if we keep it simple and we disable the document selector field also when `itemPastLoans > 0`? Then we add a similar message on the select to explain why it cannot be changed.

I am worried that the will anyway proceed, and then it can really create inconsistency around :)
This will also simply the form and avoid the modal.

src/lib/pages/backoffice/Item/ItemForms/ItemEditor/ItemForm/ItemForm.js
```diff
@@ -114,6 +135,8 @@ export class ItemForm extends Component {
         successSubmitMessage={successSubmitMessage}
         title={title}
         pid={pid}
+        modal={hasModal ? modal : null}
```
ntarocco, line 138: If we go with this, I will then do something different:
```
BaseForm
  onSubmit=() => modal and then call the original on submit if yes
```
basically, I am not sure we want to implement a modal in the base form... only if it can be reused?

Or, if we implement the modal there, then I would specialize it:
```
<BaseForm
    onSubmitCreateModal
    onSubmitEditModal
```
but then the conditions can differ.

src/lib/pages/backoffice/Item/ItemForms/ItemEditor/index.js
```diff
@@ -1 +1,17 @@
...
+
+import { fetchPastLoans } from '../../ItemDetails/ItemPastLoans/state/actions';
+import ItemEditorComponent from './ItemEditor';
+
+const mapStateToProps = state => ({
+  itemPastLoans: state.itemPastLoans.data.total,
+});
+
+const mapDispatchToProps = dispatch => ({
+  fetchPastLoans: itemPid => dispatch(fetchPastLoans(itemPid)),
```
ntarocco, line 11: I am a bit worried by adding this, yet another search, but I guess we don't have much options...
Anyway, it will happen quite rarely...
Make sure that this search finishes before rendering the form, it is blocking, maybe?

src/lib/pages/backoffice/Item/ItemDetails/ItemActionMenu/ItemActionMenu.js
jrcastro2: Yes, we can have the button disabled and make the Popup appear when we hover over. I thought it could be more user-friendly this way but I agree with you, thanks!

src/lib/pages/backoffice/Item/ItemForms/ItemEditor/ItemForm/ItemForm.js
jrcastro2: Absolutley, if that's an option would simplify things a lot. I didn't consider this option because I thought that the change of the document field in the item form is a "requirement", if it's not, I would go for this option.

src/lib/pages/backoffice/Item/ItemForms/ItemEditor/ItemForm/ItemForm.js
jrcastro2: Yes, it is reusable. Any component that uses the `BaseForm` component will be able to use the modal by passing 2 props:

1. **modal:** It's an obejct that contains the header and the content of the modal, f.e:
```
const modal = {
      header: 'Possible inconsistency',
      content:
        'You are about to change the referred document of this item. This item already has past loans that belongs to the current document. Be aware that changing the document can provoke some inconsistencies. Do you want to proceed ?',
    };
```
2. **shouldModalDisplay:**  This prop allows to pass a function to add some conditions if needed for displaying the modal. If no conditions are needed, just return true

src/lib/pages/backoffice/Item/ItemForms/ItemEditor/index.js
jrcastro2: This search will happen any time we open the form edit.
Now it's updated to wait for the search to finish before rendering the form, thanks!

src/lib/pages/backoffice/Item/ItemForms/ItemEditor/ItemForm/ItemForm.js
kprzerwa: I agree, lets block changing the item's document completely if there are loans

jrcastro2:
Removed the modal and disabled the document field if there are past loans present, an informative message is also displayed.

ntarocco:
After discussion, it would be great to allow edition of an item, blocking the `document_pid` if past or active loans...
The backend should be fixed according, see [here](https://github.com/inveniosoftware/invenio-app-ils/blob/b7d433c34cdb60b8320a015990a96891b61f0062/invenio_app_ils/items/api.py#L59)

jrcastro2:
Question for reviewers: Since we raise an error from the backend (see [here](https://github.com/inveniosoftware/invenio-app-ils/pull/989)) when updating the document if there are past loans, does it make sense to keep it only server side or should we  disable the field and add the error message ?

src/lib/pages/backoffice/Item/ItemForms/ItemEditor/ItemForm/ItemForm.js
```diff
@@ -99,8 +99,12 @@ export class ItemForm extends Component {
       successSubmitMessage,
       title,
       pid,
+      itemPastLoans,
     } = this.props;
     const initialValues = data ? this.prepareData(metadata) : {};
+    const hasPastLoans = itemPastLoans > 0;
+    const message =
+      'Not possible to edit the document field if the item already has past loans.';
```
ntarocco, line 107: Very very very minor:
```suggestion
      'Change of document is disabled because this item has active or past loans.';
```

ntarocco (commented):
Is it checking for active loans too?

jrcastro2:
> Is it checking for active loans too?

Now yes, thanks for pointing it out.

src/lib/pages/backoffice/Item/ItemForms/ItemEditor/ItemForm/ItemForm.js
```diff
@@ -99,8 +99,15 @@ export class ItemForm extends Component {
       successSubmitMessage,
       title,
       pid,
+      itemPastLoans,
     } = this.props;
     const initialValues = data ? this.prepareData(metadata) : {};
+    const hasPastOrActiveLoans =
+      itemPastLoans > 0 ||
+      (data.metadata.circulation &&
+        data.metadata.circulation.state === 'ITEM_ON_LOAN');
```
ntarocco, line 108: Really sorry to comment again @jrcastro2 :( but it is maybe safer to do:
```suggestion
        invenioConfig.CIRCULATION.loanActiveStates.includes(data.metadata.circulation.state);
```

FlorianCassayre (approved):
Tested, looks good to me.
+1 for adding the default props in `BaseForm`.

kprzerwa (approved):

PR #291, opened 2020-11-11
RSK: integrates latest changes

jrcastro2:
* Adds defaultSortingOnEmptyQueryString prop
  to ReactSearchKit component

Adds missing sort filters. 
Updates the behaviour of RSK, now the `initialState` will always be set to "Most relevant" (`bestmatch`) if it exists otherwise the first sort type will be used. For the `defaultSortingOnEmptyQueryString` will always be "Recently added" except for the patrons (because they only allow "bestmatch").

Behaviour of `defaultSortingOnEmptyQueryString` prop tested and it's correct

### DIFF
--- a/src/lib/modules/Series/SeriesLiteratureSearch/SeriesLiteratureSearch.js
+++ b/src/lib/modules/Series/SeriesLiteratureSearch/SeriesLiteratureSearch.js
@@ -1,5 +1,9 @@
 import { literatureApi } from '@api/literature';
-import { invenioConfig, setReactSearchKitInitialQueryState } from '@config';
+import {
+  invenioConfig,
+  setReactSearchKitInitialQueryState,
+  setReactSearchKitDefaultSortingOnEmptyQueryString,
+} from '@config';
 import history from '@history';
 import { LiteratureSearchOverridesMap } from '@modules/Literature/LiteratureSearchOverrides';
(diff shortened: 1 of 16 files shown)


kprzerwa (approved):

src/lib/modules/Series/SeriesLiteratureSearch/SeriesLiteratureSearch.js
```diff
@@ -1,5 +1,9 @@
 import { literatureApi } from '@api/literature';
-import { invenioConfig, setReactSearchKitInitialQueryState } from '@config';
+import {
+  invenioConfig,
+  setReactSearchKitInitialQueryState,
+  setReactSearchKitDefaultSortingOnEmptyQueryString,
```
topless, line 5: Not a showstopper but this variable name reached the 50 characters, maybe replace `ReactSearchKit` with `RSK`

ntarocco (approved):

topless (approved):

src/lib/modules/Series/SeriesLiteratureSearch/SeriesLiteratureSearch.js
kprzerwa: it's a long 🐫  :)

PR #298, opened 2020-11-12
covers: Removes cover placeholders

jrcastro2:
* Removes Placeholder component
  when no url present

image

### DIFF
--- a/src/lib/modules/Literature/LiteratureCover.js
+++ b/src/lib/modules/Literature/LiteratureCover.js
@@ -1,7 +1,7 @@
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 import Overridable from 'react-overridable';
-import { Image, Item, Placeholder } from 'semantic-ui-react';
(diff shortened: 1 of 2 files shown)


src/lib/modules/Literature/LiteratureCover.js
```diff
@@ -28,7 +29,10 @@ class LiteratureCover extends Component {
     } = this.props;
     const Cmp = asItem ? Item.Image : Image;
     const link = linkTo ? { as: Link, to: linkTo } : {};
-    return url ? (
+    const finalUrl = url
+      ? url
+      : invenioConfig.APP.INVENIO_UI_URL + '/api/static/images/placeholder.png';
```
kprzerwa, line 34: from what I remember it is done automatically by the backend, and the backend is responsible for adding the right placeholder. We shouldn't hardcode the placeholders, it is on the future developer to decide what to display

ntarocco (approved):

PR #300, opened 2020-11-13
search: handle malformed urls

### LINKED ISSUE #128: search: malformed queries raise unhandled exceptions
A minimal example of such a query: https://localhost:3000/search?f=

image

FlorianCassayre:
* updated `react-searchkit` to `1.0.0-alpha.10`
* completes inveniosoftware/react-searchkit#161, using our own config
* filter out invalid page sizes
* `filters` ~and `sortBy`~ do not need to be validated more than what is done by `UrlParamValidator`. ~That's because the sorting orders listed in the config are there to populate dropdowns, and should not be used a whitelist~ (unless we want to register all allowed fields on the frontend, which would allow us to correctly handle the labels). Similarly, most of the filters are sent by the server and are not available in the config
  * note: using an nonexistent filter/sorting won't break the UI (nor the server)
* minor refactoring in the search components
* closes #128


FlorianCassayre:
Added the validation for `sortBy` values. Also noticed that the `Sort` component does not use the url parameters to set its initial value; this bug may have been introduced in the previous release (not this one). A ticket will be opened there.

kprzerwa (approved):

PR #301, opened 2020-11-13
backoffice: Fixes internal location form bug

jrcastro2:
- Fixes internal location form error, when creating or editing it was displaying an error.
- If an ACQ order has status != CANCELLED and does not have a total price (which could be possible because is not a required field) it breaks when accesing the ACQ order details page. Added a check to display the value if it exists.
- If we edit an ACQ with order lines, the document and patron (of the order lines) will not appear even if they already exist and they do exist because they are required fields. Fixed by updating the `data.metadata.order_lines` with the values of `data.metadata.resolved_order_lines`. This makes me wonder why do we have `resolved_order_lines` and `order_lines` ?

### DIFF
--- a/src/lib/pages/backoffice/Acquisition/Order/OrderDetails/OrderStatistics.js
+++ b/src/lib/pages/backoffice/Acquisition/Order/OrderDetails/OrderStatistics.js
@@ -91,10 +91,14 @@ export class OrderStatistics extends React.Component {
     return (
       <Statistic>
         <Statistic.Label>Total</Statistic.Label>
-        <Statistic.Value>
-          {formatPrice(order.grand_total, false)}
-        </Statistic.Value>
-        <Statistic.Label>{order.grand_total.currency}</Statistic.Label>
+        {order.grand_total ? (
+          <>
+            <Statistic.Value>
+              {formatPrice(order.grand_total, false)}
+            </Statistic.Value>
+            <Statistic.Label>{order.grand_total.currency}</Statistic.Label>
+          </>
+        ) : null}
       </Statistic>
     );
   }
--- a/src/lib/pages/backoffice/Acquisition/Order/OrderEditor/OrderForm/OrderForm.js
+++ b/src/lib/pages/backoffice/Acquisition/Order/OrderEditor/OrderForm/OrderForm.js
@@ -114,7 +114,6 @@ export class OrderForm extends Component {
 
   render() {
     const { successSubmitMessage, data, title, pid, isCreate } = this.props;
-
     return (
       <BaseForm
         initialValues={data ? data.metadata : this.getDefaultValues()}
--- a/src/lib/pages/backoffice/Location/LocationForms/InternalLocationEditor/components/InternalLocationForm/InternalLocationForm.js
+++ b/src/lib/pages/backoffice/Location/LocationForms/InternalLocationEditor/components/InternalLocationForm/InternalLocationForm.js
@@ -18,7 +18,7 @@ import { Header, Segment } from 'semantic-ui-react';
 import * as Yup from 'yup';
 
 const InternalLocationSchema = Yup.object().shape({
-  location: Yup.object('missing').shape({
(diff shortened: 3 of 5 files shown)


FlorianCassayre (approved):

src/lib/pages/backoffice/Acquisition/Order/OrderDetails/OrderStatistics.js
```diff
@@ -91,10 +91,14 @@ export class OrderStatistics extends React.Component {
     return (
       <Statistic>
         <Statistic.Label>Total</Statistic.Label>
-        <Statistic.Value>
-          {formatPrice(order.grand_total, false)}
-        </Statistic.Value>
-        <Statistic.Label>{order.grand_total.currency}</Statistic.Label>
+        {order.grand_total ? (
```
topless, line 94: If I am not wrong this will display a label `Total` without any element in it, would it make sense to render the `<Statistic>` element only if there is a grand.total_value? Another approach would be to display the `<Statistic>` element and if it doesn't have a value in grand_total to display a `-` or in statistic value and currency, this will also work as expected with the styles

src/lib/pages/backoffice/Acquisition/Order/OrderEditor/OrderForm/OrderForm.js
```diff
@@ -114,7 +114,7 @@ export class OrderForm extends Component {
 
   render() {
     const { successSubmitMessage, data, title, pid, isCreate } = this.props;
-
+    data.metadata.order_lines = data.metadata.resolved_order_lines;
```
topless, line 117: Why do we re-assign a renamed value in `data.metadata` would something like the following work?

```suggestion
    const { resolved_order_lines: order_lines } = data.metadata;
```

src/lib/pages/backoffice/Location/LocationForms/InternalLocationEditor/components/InternalLocationForm/InternalLocationForm.js
```diff
@@ -18,7 +18,7 @@ import { Header, Segment } from 'semantic-ui-react';
 import * as Yup from 'yup';
 
 const InternalLocationSchema = Yup.object().shape({
-  location: Yup.object('missing').shape({
+  location: Yup.object().shape({
```
topless, line 21: Will this work for newly created internal locations which they don't have a pid yet?

src/lib/pages/backoffice/Acquisition/Order/OrderEditor/OrderForm/OrderForm.js
kprzerwa: I think he is assigning actually here, not renaming (replacing existing `order_lines` with `resolved_order_lines` to display resolved patron and document in the form)
which I think it is a bit confusing in general, we are discussing on chat

src/lib/pages/backoffice/Acquisition/Order/OrderDetails/OrderStatistics.js
jrcastro2: Hmm, I kept it like this because it's harmonized with the rest of the page, if we remove the `<Statistic>` component then nothing will appear at the top right of the image. 
I agree with you to add a default value for empty fields but I think we should update the rest of the fields too to prevent different "styles" in the same page.
image

src/lib/pages/backoffice/Location/LocationForms/InternalLocationEditor/components/InternalLocationForm/InternalLocationForm.js
jrcastro2: Yes, currently if you try to create or edit an internal location in `dev` it stays in loading state for ever. This should fix it. (I thought it was a typo, does that "missing" means something ? )

ntarocco (approved):

topless (approved):

PR #302, opened 2020-11-13
Routes: add overridable route

KonstantinaStoikou:
- adds overridable routes in frontsite
- exports necessary components to be used in cds-ils ([pr](https://github.com/CERNDocumentServer/cds-ils/pull/228))

### DIFF
--- a/src/lib/routes/frontsite/Frontsite.js
+++ b/src/lib/routes/frontsite/Frontsite.js
@@ -18,8 +18,9 @@ import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 import { Route, Switch } from 'react-router-dom';
 import { Container } from 'semantic-ui-react';
+import Overridable from 'react-overridable';
 
-export default class FrontSite extends Component {
+export class FrontSite extends Component {
   renderCustomStaticPages = () => {
     const { customStaticPages } = this.props;
     customStaticPages();
@@ -74,6 +75,8 @@ export default class FrontSite extends Component {
               <Route key={route} exact path={route} component={StaticPage} />
             ))}
             {this.renderCustomStaticPages()}
+
+            <Overridable id="Frontsite.route" {...this.props} />
             <Route>
               <NotFound />
             </Route>
@@ -92,3 +95,5 @@ FrontSite.propTypes = {
 FrontSite.defaultProps = {
   customStaticPages: () => {},
 };
+
+export default Overridable.component('Frontsite', FrontSite);
--- a/src/lib/index.js
+++ b/src/lib/index.js
@@ -30,3 +30,6 @@ export { FrontSiteRoutes } from '@routes/frontsite/frontsiteUrls';
 export { default as InvenioILSApp } from './App';
 export { default as history } from './history';
 export { default as store } from './store';
+export { http } from '@api/base';
+export { NotFound } from '@components/HttpErrors';
+export { withCancel, recordToPidType } from '@api/utils';


src/lib/routes/frontsite/Frontsite.js
```diff
@@ -28,59 +29,67 @@ export default class FrontSite extends Component {
...
+              <Route
+                exact
+                path={FrontSiteRoutes.errors}
+                component={ErrorsPage}
+              />
+              {staticPagesRoutes.map(route => (
+                <Route key={route} exact path={route} component={StaticPage} />
+              ))}
+              {this.renderCustomStaticPages()}
+              <Overridable id="Frontsite.route" {...this.props} />
```
kprzerwa, line 83: did you check if it does not break the other routes ?

src/lib/routes/frontsite/Frontsite.js
```diff
@@ -28,59 +29,67 @@ export default class FrontSite extends Component {
...
-            ))}
-            {this.renderCustomStaticPages()}
-            <Route>
-              <NotFound />
-            </Route>
-          </Switch>
-        </Container>
-        <ILSFooter />
-      </div>
+      <Overridable id="Frontsite.layout" {...this.props}>
```
kprzerwa, line 32: I'm not sure if it is needed to overwrite all the routes at this point
@ntarocco @topless @jrcastro2 @FlorianCassayre what do you guys think?

ntarocco: The biggest issue I see here is that we need to think that Overridable is a black box: when overriding it, we cannot access to the outside props/vars/imports.
So we should pass any extra props, e.g. the `staticPagesRoutes`.

We should probably take the habit to extract the inner part in another cmp, even if we could still do this mistake:

```
<Overridable ... staticPagesRoutes>
  <NewCmp props... staticPagesRoutes>
</Overridable>

class NewCmp ...
    staticPagesRoutes
```

src/lib/pages/frontsite/Documents/DocumentDetails/DocumentItems/index.js
```diff
@@ -1,8 +1,8 @@
 import { connect } from 'react-redux';
 import DocumentItemsComponent from './DocumentItems';
 
-const mapStateToProps = state => ({
-  documentDetails: state.documentDetailsFront.data,
+const mapStateToProps = (state, ownProps) => ({
+  documentDetails: ownProps.documentDetails || state.documentDetailsFront.data,
```
kprzerwa, line 5: did you check if it does not break the DocumentDetails page of standalone `react-invenio-app-ils`?

src/lib/pages/frontsite/Documents/DocumentDetails/index.js
```diff
@@ -7,9 +7,11 @@ const mapDispatchToProps = dispatch => ({
     dispatch(fetchDocumentsDetails(documentPid, { fetchLoansInfo: true })),
 });
 
-const mapStateToProps = state => ({
-  isLoading: state.documentDetailsFront.isLoading,
-  documentDetails: state.documentDetailsFront.data,
+const mapStateToProps = (state, ownProps) => ({
+  isLoading: ownProps.documentDetails
+    ? false
```
kprzerwa, line 12: the redux initial state might not be applied because of this

src/lib/pages/frontsite/Documents/DocumentDetails/DocumentDetails.js
```diff
@@ -126,8 +126,10 @@ DocumentDetailsLayout.defaultProps = {
 
 class DocumentDetails extends Component {
   componentDidMount() {
-    const { match } = this.props;
-    this.fetchDocumentsDetails(match.params.documentPid);
+    const { documentDetails, match } = this.props;
+    if (Object.keys(documentDetails.metadata).length === 0) {
```
kprzerwa, line 130: it would be much easier to just export (out of this library) the version of this component without redux mapping to be able to to use it outside, none of these changes would be necessary then.

src/lib/pages/frontsite/Documents/DocumentDetails/DocumentItems/index.js
KonstantinaStoikou: I checked, it doesn't break it.

src/lib/pages/frontsite/Documents/DocumentDetails/DocumentDetails.js
KonstantinaStoikou: I still need to keep some of the changes otherwise it is not working.

kprzerwa, line 130: nit: I would suggest to use `isEmpty` of `lodash` instead

kprzerwa (approved):

src/lib/pages/frontsite/Documents/DocumentDetails/DocumentDetails.js
```diff
@@ -126,20 +126,25 @@ DocumentDetailsLayout.defaultProps = {
 
 class DocumentDetails extends Component {
   componentDidMount() {
-    const { match } = this.props;
-    this.fetchDocumentsDetails(match.params.documentPid);
+    const { documentDetails, match } = this.props;
+    if (_isEmpty(documentDetails.metadata)) {
```
ntarocco, line 130: In which cases the document metadata is already provided via a prop?
Is it happening always?

kprzerwa: we exported this component as not redux controlled, to be able to use it independently. In this case the data will be passed from ownProps instead of the redux mapping. We are using it to save one frontend redirect when trying to access record via legacy_recid.

kprzerwa (approved):

PR #305, opened 2020-11-20
backoffice: fix the workflow of LiteratureRequest

### LINKED ISSUE #290: backoffice: literature request workflow
* on the "Select provider" step:
  * both searchbars do not include the searched text in the query url (add a method in the query builder)
  * when clicking on a result of the acquisition order searchbar, it should bind the record to the request and proceed to the next step (as for the ILL loan). Currently it acts as an `ESSelector` form field -> this happens when trying to add a provider in an already accepted request because it throws a server error and hangs (because demo requests from cli don't have a provider assigned to them)
  * the user is asked to select an existing order/loan or to create a new one. The creation button does not automatically link the newly created record to the request (the button is merely a link to the corresponding creation form)
    * what should happen: the librarian clicks on either create create button, fill the form, submit it, should be redirected to the request details page and should observe that the request has progressed to the next step
    * alternatively, open the creation form in a new page so that the user can come back to their request (less good but easier to implement)

See also inveniosoftware/invenio-app-ils#973

KonstantinaStoikou:
- redirect order and borrowing request forms to literature request form
- add function `listWithPendingStatus` to be used in Provider's ESSelector because previously it displayed all results with `status=PENDING` without taking into account the search query
Closes #290

### DIFF
--- a/src/lib/api/acquisition/order.js
+++ b/src/lib/api/acquisition/order.js
@@ -1,6 +1,7 @@
 import { http, apiConfig } from '@api/base';
 import { prepareSumQuery } from '@api/utils';
 import { orderSerializer as serializer } from './serializers';
+import { invenioConfig } from '@config';
 
 const orderURL = '/acquisition/orders/';
 
@@ -31,6 +32,20 @@ const list = async query => {
   return response;
 };
 
+const listWithPendingStatus = async query => {
+  const searchCriteria = orderApi
+    .query()
+    .withState(`(${invenioConfig.ACQ_ORDERS.pendingStatuses.join(' OR ')})`)
+    .withQuery(query)
+    .qs();
+  const response = await http.get(`${orderURL}?q=${searchCriteria}`);
+  response.data.total = response.data.hits.total;
+  response.data.hits = response.data.hits.hits.map(hit =>
+    serializer.fromJSON(hit)
+  );
+  return response;
+};
+
 const count = async query => {
   const response = await http.get(`${orderURL}?q=${query}`);
   response.data = response.data.hits.total;
@@ -43,6 +58,7 @@ class QueryBuilder {
     this.recipientQuery = [];
     this.vendorQuery = [];
     this.vendorPidQuery = [];
+    this.query = [];
     this.sortByQuery = '';
     this.stateQuery = [];
   }
@@ -87,6 +103,14 @@ class QueryBuilder {
     return this;
   }
 
+  withQuery(query) {
+    if (!query) {
+      throw TypeError('Query argument missing');
+    }
+    this.query.push(`${query}`);
+    return this;
+  }
+
   sortBy(order = 'bestmatch') {
     this.sortByQuery = `&sort=${order}`;
     return this;
@@ -96,6 +120,7 @@ class QueryBuilder {
     const searchCriteria = this.patronQuery
       .concat(this.recipientQuery, this.vendorQuery, this.vendorPidQuery)
       .concat(this.stateQuery)
+      .concat(this.query)
       .join(' AND ');
     return `${searchCriteria}${this.sortByQuery}`;
   }
@@ -111,6 +136,7 @@ export const orderApi = {
   create: create,
   update: update,
   list: list,
+  listWithPendingStatus: listWithPendingStatus,
   count: count,
   query: queryBuilder,
   serializer: serializer,
--- a/src/lib/pages/backoffice/Acquisition/Order/OrderEditor/OrderEditor.js
+++ b/src/lib/pages/backoffice/Acquisition/Order/OrderEditor/OrderEditor.js
@@ -65,7 +65,7 @@ export class OrderEditor extends Component {
     const request = _get(this.props, 'location.state', null);
     if (!request) return null;
     return {
-      documentRequestPid: request.pid,
+      documentRequestPid: request.metadata.pid,
       metadata: {
         order_lines: [
           {
--- a/src/lib/pages/backoffice/Acquisition/Order/OrderEditor/OrderForm/OrderForm.js
+++ b/src/lib/pages/backoffice/Acquisition/Order/OrderEditor/OrderForm/OrderForm.js
@@ -13,6 +13,10 @@ import { OrderInfo } from './OrderInfo';
 import { OrderLines } from './OrderLines';
 import { Payment } from './Payment';
 import * as Yup from 'yup';
+import _get from 'lodash/get';
+import { documentRequestApi } from '@api/documentRequests';
+import { BackOfficeRoutes } from '@routes/urls';
+import { invenioConfig } from '@config';
 
 const orderSubmitSerializer = values => {
   const submitValues = { ...values };
@@ -100,10 +104,25 @@ export class OrderForm extends Component {
     return orderApi.update(pid, data);
   };
 
-  successCallback = response => {
-    goTo(
-      AcquisitionRoutes.orderDetailsFor(getIn(response, 'data.metadata.pid'))
+  successCallback = async response => {
+    const order = getIn(response, 'data');
+    const documentRequestPid = _get(
+      this.props,
+      'data.documentRequestPid',
+      null
     );
+    if (documentRequestPid) {
+      await documentRequestApi.addProvider(documentRequestPid, {
+        physical_item_provider: {
+          pid: order.metadata.pid,
+          pid_type:
+            invenioConfig.DOCUMENT_REQUESTS.physicalItemProviders.acq.pid_type,
+        },
+      });
+      goTo(BackOfficeRoutes.documentRequestDetailsFor(documentRequestPid));
+    } else {
+      goTo(AcquisitionRoutes.orderDetailsFor(order.metadata.pid));
+    }
   };
 
   getDefaultValues() {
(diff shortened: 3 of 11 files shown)


src/lib/api/acquisition/order.js
```diff
@@ -111,6 +120,7 @@ export const orderApi = {
   create: create,
   update: update,
   list: list,
+  listWithPendingStatus: listWithPendingStatus,
```
FlorianCassayre, line 123: (minor) I'm in favour of using the request builder for that; say `withQuery` that allows to enter a free text query.

src/lib/pages/backoffice/Acquisition/Order/OrderEditor/OrderEditor.js
```diff
@@ -65,7 +65,7 @@ export class OrderEditor extends Component {
     const request = _get(this.props, 'location.state', null);
     if (!request) return null;
     return {
-      documentRequestPid: request.pid,
+      documentRequestPid: request.id,
```
kprzerwa, line 68: did you mean `request.metadata.pid`? `.id` is inconsistent with all the other record types we have

FlorianCassayre, line 68: Question: what does it change?

src/lib/pages/backoffice/Acquisition/Order/OrderEditor/OrderForm/OrderForm.js
```diff
@@ -100,10 +104,25 @@ export class OrderForm extends Component {
...
+    const order = getIn(response, 'data');
+    const documentRequestPid = _get(
+      this.props,
+      'data.documentRequestPid',
+      null
     );
+    if (documentRequestPid) {
+      await documentRequestApi.addProvider(documentRequestPid, {
+        physical_item_provider: {
+          pid: order.pid,
```
kprzerwa, line 117: `order.metadata.pid` (verify if it works please), we try to abandon the `record.pid` because it is inconsistent with the objects structures, we introduced it at some point in the serializers, but it is confusing IMO

src/lib/pages/backoffice/Document/DocumentForms/DocumentEditor/DocumentEditor.js
```diff
@@ -37,7 +37,7 @@ export class DocumentEditor extends Component {
     const request = _get(this.props, 'location.state', null);
     if (!request) return null;
     return {
-      documentRequestPid: request.pid,
+      documentRequestPid: request.id,
```
kprzerwa, line 40: `documentRequestPid: request.metadata.pid`, as above, to be verified if it works

src/lib/pages/backoffice/DocumentRequest/DocumentRequestDetails/DocumentRequestSteps/ProviderStep/ProviderStep.js
```diff
@@ -82,14 +82,7 @@ class AcqProvider extends Component {
...
               onSelectResult={this.onSelectResult}
-              query={() =>
-                orderApi.list(
-                  orderApi
-                    .query()
-                    .withState('PENDING')
-                    .qs()
-                )
-              }
+              query={orderApi.listWithPendingStatus}
```
kprzerwa, line 85: I'm not sure if supplying only the `PENDING` ones will be OK in this case. @topless @ntarocco is there any use case we would like to connect orders/ILLs in the states different than `PENDING`?

src/lib/pages/backoffice/ILL/BorrowingRequest/BorrowingRequestEditor/BorrowingRequestEditor.js
```diff
@@ -36,6 +36,7 @@ export class BorrowingRequestEditor extends Component {
     } = this.props;
     if (!request) return null;
     return {
+      documentRequestPid: request.id,
```
kprzerwa, line 39: `request.metadata.pid` as in the comments above

src/lib/pages/backoffice/Acquisition/Order/OrderEditor/OrderForm/OrderForm.js
```diff
@@ -100,10 +104,25 @@ export class OrderForm extends Component {
     return orderApi.update(pid, data);
   };
 
-  successCallback = response => {
-    goTo(
-      AcquisitionRoutes.orderDetailsFor(getIn(response, 'data.metadata.pid'))
+  successCallback = async response => {
```
FlorianCassayre, line 107: Good job, great UX design.
That said I'm not 100% sure about making the request in the form submit callback. For instance, how should we handle exception? (currently not implemented as fair as I'm aware)

src/lib/api/acquisition/order.js
KonstantinaStoikou: It wouldn't work with `withQuery` because we would need to pass arguments in the call here https://github.com/inveniosoftware/react-invenio-app-ils/blob/7574e2d949f18d414100d1dfcecc5b1e4bd6dd86/src/lib/pages/backoffice/DocumentRequest/DocumentRequestDetails/DocumentRequestSteps/ProviderStep/ProviderStep.js#L85
and in `HitsSearch.js` we need to pass only the reference to the function if we want the query string to be taken into account https://github.com/inveniosoftware/react-invenio-app-ils/blob/64754f29ac5dcf7876b4d06095e8cca76379cc72/src/lib/modules/ESSelector/HitsSearch.js#L81

FlorianCassayre: Yes, in this case you can always create an anonymous function:
```javascript
query={
  q => orderApi
        .query()
        .withStatus('PENDING')
        .withQuery(q) // Pass the argument
        .qs()
}
```
Or, better create a function in the class and reference it.

KonstantinaStoikou: The query builder will build the query (`queryString` + `SearchCriteria`) but we still need to pass it as an argument in the `list` function to create the url and make the request.

src/lib/pages/backoffice/Acquisition/Order/OrderEditor/OrderEditor.js
KonstantinaStoikou: I intentionally changed it to `id` because yesterday I tested it and the `pid` was sometimes undefined but the `id` had a value but I probably did something wrong because now that I tested it it's working fine.

KonstantinaStoikou: Answered above.

src/lib/pages/backoffice/Acquisition/Order/OrderEditor/OrderForm/OrderForm.js
KonstantinaStoikou: If you are talking about exceptions that might occur during adding the provider, I think we can add an action to dispatch an error message instead of directly calling the `addProvider` from the api in the callback.

src/lib/api/acquisition/order.js
FlorianCassayre: Sorry you're right I forgot it returned an url instead of an http request.

kprzerwa: you could still use the query builder to build the url in borrowingRequest.js instead of hardcoding it

src/lib/pages/backoffice/Acquisition/Order/OrderEditor/OrderForm/OrderForm.js
ntarocco: Question for reviewers: I am wondering if this should be done on the backend. After a record creation (signal), update the document request...

src/lib/pages/backoffice/DocumentRequest/DocumentRequestDetails/DocumentRequestSteps/ProviderStep/ProviderStep.js
ntarocco: Maybe it will be safer to get the orders that do not have yet a `document_request_pid`, independently of the state.
Imagine the scenario where the librarian will not follow the workflow and:
1. see the doc request
2. go to the order page and create a new order
3. come back to the doc request and want to attach it.

WDYT?

src/lib/pages/backoffice/Acquisition/Order/OrderEditor/OrderForm/OrderForm.js
kprzerwa: I'm not sure why it was not done in the backend in the first place. at this point I would say to ticketize it for refatoring @ntarocco WDYT?

ntarocco: It is harder than I thought. Let's keep it like this.

kprzerwa (approved):

src/lib/api/acquisition/order.js
```diff
@@ -31,6 +32,22 @@ const list = async query => {
   return response;
 };
 
+const listWithPendingStatus = async query => {
+  const searchCriteria = orderApi
+    .query()
+    .withState(
+      `(${invenioConfig.ILL_BORROWING_REQUESTS.pendingStatuses.join(' OR ')})`
```
kprzerwa, line 39: 👀  please change to acq states

src/lib/pages/backoffice/DocumentRequest/DocumentRequestDetails/DocumentRequestSteps/ProviderStep/ProviderStep.js
KonstantinaStoikou: There is no `document_request_pid` in the order and there is no connection between the order and the document request from the order's side.

PR #313, opened 2020-12-15
Routes: fix not-found route bug

### LINKED ISSUE #312: routes: broken static routes and 404
As a side-effect of the overridable routes feature (#302):
- the static routes are not working anymore and default to a blank ILS page instead (that means for instance that the favicon cannot be served anymore)
- same thing for invalid routes; they should be displaying the 404 layout

KonstantinaStoikou:
Closes #312 
- removed `homepage` from package.json because it was setting the static resources url to `https://127.0.0.1/inveniosoftware/react-invenio-app-ils/<static_resource>` instead of `https://127.0.0.1/<static_resource>`
- wrapped Overridable route with NotFound route because before the fallback route was the Overridable instead if the NotFound

### DIFF
--- a/package.json
+++ b/package.json
@@ -116,7 +116,6 @@
     "name": "CERN",
     "email": "info@inveniosoftware.org"
   },
-  "homepage": "https://github.com/inveniosoftware/react-invenio-app-ils/",
   "bugs": {
     "url": "https://github.com/inveniosoftware/react-invenio-app-ils/issues"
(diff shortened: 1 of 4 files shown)


FlorianCassayre (approved):
Thanks so much for fixing the annoying thing with `homepage` 😅 👍

kprzerwa (approved):

PR #317, opened 2021-01-04
covers: alignment on frontsite search

### LINKED ISSUE #296: Covers improvements
### Search page:
- provide one size for each cover. In case of the cover is smaller than the container fixed width and height, leave a white background around the cover, when it is bigger display with `background-size: cover` (or apply for all if the first variant too difficult)

### Cover styling (details page and search)
- apply a discrete, (but visible) border around the cover image, since some of the covers have white background and it is not possible to distinguish the edge of the image)

FlorianCassayre:
Make the cover frame constant, fit the image in it while preserving the aspect ratio and avoiding overflow. The images are always centred.

Closes #296

Notes:
* this does not apply on mobile, but it applies on the computer stacked view
* I didn't find a way to inherit the height or make it relative to the width, so I hardcoded it (`200px`)
* there is no left/right margins so the wide covers will match the overlay width (typically square or landscape images). I personally don't find it odd but it can be changed if necessary

image

[A screenshot of the stacked view](https://user-images.githubusercontent.com/9027075/103626245-358d0b00-4f3c-11eb-8a14-a2e51b1cbc4f.png)

### DIFF
--- a/src/semantic-ui/site/elements/image.overrides
+++ b/src/semantic-ui/site/elements/image.overrides
@@ -5,3 +5,18 @@ Image Overrides - REACT-INVENIO-APP-ILS
 .document-panel .ui.image.large{
     width: 350px;
 }
+
+.ui.grid.grid-documents-search {
+    .ui.image.image-cover {
+         object-fit: contain;
+         width: 100%;
+         height: 200px;
+         background: none !important;
+    }
+}
+
+.ui {
+    a > .image.image-cover, .image.image-cover {
+         filter: drop-shadow(0px 1px 0px #eee) drop-shadow(1px 0px 0px #eee) drop-shadow(0px -1px 0px #eee) drop-shadow(-1px 0px 0px #eee);
+    }
+}
--- a/src/lib/modules/Literature/LiteratureCover.js
+++ b/src/lib/modules/Literature/LiteratureCover.js
@@ -39,6 +39,7 @@ class LiteratureCover extends Component {
             onError={e => (e.target.style.display = 'none')}
             src={url}
             size={size}
+            className="image-cover"
             {...uiProps}
           />
         )}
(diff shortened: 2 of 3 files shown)


src/semantic-ui/site/elements/image.overrides
```diff
@@ -5,3 +5,12 @@ Image Overrides - REACT-INVENIO-APP-ILS
 .document-panel .ui.image.large{
     width: 350px;
 }
+
+.ui.grid.grid-documents-search {
+    .ui.image {
+         object-fit: contain;
```
ntarocco, line 11: it will definitively not work on IE11! :) :D

ntarocco (approved):

kprzerwa (approved):

kprzerwa (commented):
Screenshot 2021-01-06 at 15 01 42
If we go with this solution, we should change the placeholder image too, because it does not look good - it takes up more space than the covers and attracts more attention as a result
ping @ntarocco 
could we change the placeholder?

FlorianCassayre:
You are right, I was also thinking about changing this cover image. In my opinion it should be something simple that suggests a book. Here is [an example](https://i.stack.imgur.com/1hvpD.jpg) used by a book provider. It could also simply be a monochromatic rectangle with a border.
`5/8` seems like a common book format.

ntarocco:
> You are right, I was also thinking about changing this cover image. In my opinion it should be something simple that suggests a book. Here is [an example](https://i.stack.imgur.com/1hvpD.jpg) used by a book provider. It could also simply be a monochromatic rectangle with a border (like Google Books).
> `5/8` seems like a common book format.

Great idea! Go ahead!

FlorianCassayre:
Side by side comparison of the shadow effect (without / with):
Screenshot 2021-01-07 at 10 58 00

On white covers it creates a nice contrast that helps locate the border. On other covers it's not really noticeable, but it makes the borders appear less sharp (obviously).

Same comparison, this time with a solid border:
Screenshot 2021-01-07 at 14 04 51

I tend to prefer the border because it preserves the sharpness.

Now for the placeholder, after a few iterations I came up with the following design:

image

Thanks to its light color the eye is dragged towards the other covers first. The shapes are vector centred. This feature will have its dedicated PR on `invenio-app-ils` (because that's where it is defined).

Any comments welcome 🙂

jrcastro2 (approved):

PR #324, opened 2021-01-06
dependencies: minor upgrades

jrcastro2:
Some minor upgrades


ntarocco (approved):

kprzerwa (approved):
thanks for that! 
we should do the same in cds-ils

PR #328, opened 2021-01-08
authors: Improves authors display

jrcastro2:
* Now we can show all authors in the details page
* Reduced overall displayed authors limit from 10 to 5
* closes https://github.com/CERNDocumentServer/cds-ils/issues/231

In terms of performance it's slightly slower when performing a search of 15 documents with 5k authors each. 
While a usual search takes around 400-450 ms, the heavy search takes 150-200 ms longer (550-600 ms).

## Usual search
image

## Heavy search
image

----

image

image

### DIFF
--- a/src/lib/modules/Document/DocumentAuthors.js
+++ b/src/lib/modules/Document/DocumentAuthors.js
@@ -2,6 +2,7 @@ import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 import Overridable from 'react-overridable';
 import { Icon, List, Popup } from 'semantic-ui-react';
+import { invenioConfig } from '@config';
 
 const ET_AL_LABEL = 'et al.';
 
@@ -170,6 +171,7 @@ class DocumentAuthors extends Component {
       withPopUpShowMoreFields,
       showAllFieldsInPopUp,
       listItemAs,
+      withVerticalScroll,
     } = this.props;
     const { isExpanded } = this.state;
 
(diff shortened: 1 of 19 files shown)


FlorianCassayre (approved):
Good job, from this ticket I was expecting more changes for fewer features. 🙂

I can possibly suggest to wrap the magic value `5` in a config value, or to reuse the existing `LITERATURE.authors.maxDisplay` since they are all related and would make sense to be the same.

src/lib/modules/Document/DocumentAuthors.js
```diff
@@ -246,6 +255,7 @@ DocumentAuthors.defaultProps = {
   withPopUpShowMoreFields: false,
   showAllFieldsInPopUp: true,
   listItemAs: null,
+  noClassName: false,
```
ntarocco, line 258: I am not sure I understand very well the meaning of `noClassName: false`. Is it to have a default?

kprzerwa: I'm also curious

jrcastro2: Yes, most of the times this component is used it needs the className `document-authors-list-wrapper`, I added this prop for the cases where we don't want that className (only 2 cases). (Did it this way to prevent passing the same className all the time we use this component, but can change it to `className` and use it)

ntarocco: thank you for the explanation! In this case, can we maybe improve the naming of the prop?
If we use something like `withListWrapper` it might be more clear. I would choose a name that explain what it does: it looks like that is setting a max height and adding a scroll? Some options:
* `withVerticalScroll`
* `withMaxSize`
any other idea?

src/lib/modules/Document/DocumentAuthors.js
```diff
@@ -195,15 +197,21 @@ class DocumentAuthors extends Component {
...
+        >
           {prefix}
           <List horizontal className="document-authors-list" as={listItemAs}>
             {displayedAuthors.map((author, index) => {
               const isLast = index === displayedAuthors.length - 1;
               return (
                 <List.Item key={author.full_name}>
                   {author.full_name}
-                  {withPopUpShowMoreFields && (
+                  {withPopUpShowMoreFields && allAuthors.length < 5 && (
```
ntarocco, line 214: Just for the sake of naming numbers in the code, I would have this as a prop:
```suggestion
                  {withPopUpShowMoreFields && allAuthors.length < showMaxAuthorsCount && (
```
and then
```
DocumentAuthors.defaultProps = {
    ...
    showMaxAuthorsCount: 5
```
What do you think?

jrcastro2: Yes, thanks for noticing! We can use `invenioConfig.LITERATURE.authors.maxDisplay` instead

jrcastro2:
Comments fixed and improved display of authors.
image
image

kprzerwa (approved):

src/lib/modules/Document/DocumentAuthors.js
jrcastro2: `withVerticalScroll` seems good for me.

KonstantinaStoikou (approved):

PR #349, opened 2021-02-05
vocabularies: set vocab max results window to 500

KonstantinaStoikou:
- vocab max results window to 500
- max export results window to 4000

### DIFF
--- a/src/lib/config/defaultConfig.js
+++ b/src/lib/config/defaultConfig.js
@@ -14,6 +14,8 @@ export const APP_CONFIG = {
   INVENIO_UI_URL: process.env.REACT_APP_INVENIO_UI_URL,
   LOGO_SRC: process.env.PUBLIC_URL + '/logo-invenio-ils.svg',
   MAX_RESULTS_WINDOW: 10000,
+  VOCAB_MAX_RESULTS_WINDOW: 500,
+  EXPORT_MAX_RESULTS: 4000,
   OAUTH_PROVIDERS: {
     github: {
       enabled: true,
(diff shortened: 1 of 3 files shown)


kprzerwa (approved):

jrcastro2 (approved):

ntarocco (commented):
Does this solve the export issue in the end?

src/lib/config/defaultConfig.js
```diff
@@ -13,7 +13,8 @@ export const APP_CONFIG = {
   SEARCH_READY_DELAY: 2000,
   INVENIO_UI_URL: process.env.REACT_APP_INVENIO_UI_URL,
   LOGO_SRC: process.env.PUBLIC_URL + '/logo-invenio-ils.svg',
-  MAX_RESULTS_WINDOW: 10000,
+  MAX_RESULTS_WINDOW: 5000,
```
ntarocco, line 16: I would change it to the following:
```suggestion
  SEARCH_MAX_RESULTS: 10000,  // should match search `max_results_window` in the backend
  VOCABULARIES_MAX_RESULTS: 500,
  EXPORT_MAX_RESULTS: 5000,
```
This is because the backend allows 10K, and that number is used to calculate the pagination. If we keep it aligned with the backend, we can actually jump to that page with the pagination.

Then, for vocab, perfect.
For the export, since we need to export all of them at one, I would indeed introduce a new var but not change the default one.

WDYT?

KonstantinaStoikou:
> Does this solve the export issue in the end?

I tried with the 5000 and sometimes it threw `Bad Gateway`. With 4000 it seems it works.

ntarocco (approved):

PR #361, opened 2021-02-15
extensions: boolean values

jrcastro2:
* Fix display of boolean values
image

### DIFF
--- a/src/lib/components/backoffice/MetadataTable/MetadataTable.js
+++ b/src/lib/components/backoffice/MetadataTable/MetadataTable.js
@@ -1,21 +1,19 @@
 import React, { Component } from 'react';
 import { Table } from 'semantic-ui-react';
 import PropTypes from 'prop-types';
+import { prettyPrintBooleanValue } from '@components/utils';
 
 export default class MetadataTable extends Component {
   renderRows() {
     const { labelWidth, rows } = this.props;
-
     return rows.map((row) => (
       <Table.Row key={row.name}>
         <Table.Cell width={labelWidth}>
           {row.icon} {row.name}
         </Table.Cell>
         <Table.Cell>
-          {typeof row.value == 'boolean'
-            ? row.value
-              ? 'Yes'
-              : 'No'
+          {typeof row.value === 'boolean'
+            ? prettyPrintBooleanValue(row.value)
             : row.value}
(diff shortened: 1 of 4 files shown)


src/lib/modules/Document/DocumentMetadataExtensions.js
```diff
@@ -5,19 +5,24 @@ import React from 'react';
...
       <Table definition>
         <Table.Body>
           {_keys(extensions).map((key) => (
             <Table.Row key={key}>
               <Table.Cell width={4}>
-                {invenioConfig.DOCUMENTS.extensions.fields[key].label}
+                {documentExtensions.fields[key].label}
+              </Table.Cell>
+              <Table.Cell>
+                {documentExtensions.fields[key].type === 'boolean'
```
kprzerwa, line 20: since we do this, maybe it makes sense to have a global solution for the boolean values display, not only for the extensions ? ping @ntarocco

ntarocco: Yes it makes sense, e.g. an utility `prettyPrintExtensionField(type, value)`

src/lib/components/backoffice/MetadataTable/MetadataTable.js
```diff
@@ -13,9 +14,7 @@ export default class MetadataTable extends Component {
         </Table.Cell>
         <Table.Cell>
           {typeof row.value == 'boolean'
-            ? row.value
-              ? 'Yes'
-              : 'No'
+            ? prettyPrintBooleanValue(row.value)
```
ntarocco, line 17: I am not sure I understand the logic here: isn't it the type already in `row.type`? If yes I would simply:
```
{row.type === 'boolean' ? pretty... : row.value}
```
Remember to use `===` and not `==` ;)

ntarocco (commented):
LGTM! Just a  very small comment if you want to have a look.

src/lib/components/backoffice/MetadataTable/MetadataTable.js
jrcastro2: No, we don't have the `type` here, we only have it in the extensions fields.

ntarocco (approved):

PR #370, opened 2021-02-22
[DEPENDS ON IAILS#1041] document: frontsite add urls display

kprzerwa:


src/lib/modules/Literature/LiteratureUrls.js
```diff
@@ -0,0 +1,39 @@
...
+          })}
+        </>
+      </Overridable>
+    );
+  }
+}
+
+LiteratureUrls.propTypes = {
+  urlArray: PropTypes.array.isRequired,
+  separator: PropTypes.string,
```
ntarocco, line 32: String or component as a separator? Isn't it maybe better to display a list?

ntarocco (approved):

jrcastro2 (approved):

src/lib/modules/Literature/LiteratureUrls.js
kprzerwa: I tried with a list, it does not look good IMO, in terms of leaving a lot of blank space when there is many links and each has a short description

PR #371, opened 2021-02-22
[DEPENDS ON IAI #1042] documentForm: updates

jrcastro2:
* Material and subject are vocabularies now
* Moved up the fields Imprint, Identifiers
  and Alternative titles
* closes https://github.com/CERNDocumentServer/cds-ils/issues/330
* depends on https://github.com/inveniosoftware/invenio-app-ils/pull/1042

image

### DIFF
--- a/src/lib/config/common.js
+++ b/src/lib/config/common.js
@@ -2,6 +2,8 @@
 
 export const DEFAULT_CURRENCY = 'CHF';
 
+export const DEFAULT_LANGUAGE = 'ENG';
+
 // label key is used for labeling bucket values
 // text key is used for semantic ui dropdown options display
 
--- a/src/lib/config/defaultConfig.js
+++ b/src/lib/config/defaultConfig.js
@@ -1,6 +1,7 @@
 import {
   ACQ_ORDER_STATUSES,
   DEFAULT_CURRENCY,
+  DEFAULT_LANGUAGE,
   DOCUMENT_RELATIONS,
   DOCUMENT_TYPES,
   ILL_BORROWING_REQUESTS_STATUSES,
@@ -30,6 +31,7 @@ export const APP_CONFIG = {
   REST_ENDOINTS_BASE_URL: process.env.REACT_APP_INVENIO_REST_ENDPOINTS_BASE_URL,
   SUCCESS_AUTO_DISMISS_SECONDS: 10,
   DEFAULT_CURRENCY: DEFAULT_CURRENCY,
+  DEFAULT_LANGUAGE: DEFAULT_LANGUAGE,
   DEFAULT_RESULTS_SIZE: 15,
   i18n: {
     priceLocale: 'fr-CH',
@@ -911,6 +913,8 @@ export const RECORDS_CONFIG = {
       license: 'license',
       tags: 'tag',
       type: 'document_type',
+      doc_subjects: 'doc_subjects',
+      doc_identifiers_materials: 'doc_identifiers_materials',
     },
     documentRequests: {
       doc_req_type: 'doc_req_type',
--- a/src/lib/pages/backoffice/Document/DocumentDetails/DocumentMetadata/DocumentContents.js
+++ b/src/lib/pages/backoffice/Document/DocumentDetails/DocumentMetadata/DocumentContents.js
@@ -53,7 +53,7 @@ export class DocumentContents extends Component {
           {!_isEmpty(document.metadata.subjects) && (
             <>
               <Divider />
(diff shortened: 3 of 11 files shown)


src/lib/forms/core/VocabularyField.js
```diff
@@ -44,13 +44,24 @@ export class VocabularyField extends React.Component {
...
+
   fetchVocabularies = async (serializer) => {
-    const { label } = this.props;
+    const { label, orderBy } = this.props;
     this.cancellableFetchVocabularies = withCancel(this.query());
     try {
       const response = await this.cancellableFetchVocabularies.promise;
       const options = response.data.hits.map((hit) => serializer(hit));
-
+      if (orderBy) {
```
ntarocco, line 58: I understand why you have implemented like this, but hardcoding `Creative Commons` in the cmp below it means hardcoding CERN CDS behaviour in InvenioILS, and we don't want to do that.

I see 2 options her:
1. manipulate the vocab in the CDS-ILS backend and changed the order of each item, putting `CC` as first
2. introduce an `order` field in the vocab, and then put there an integer, e.g order: 0, order: 1... then, set `CC` as order: 0 and have the Elasticsearch query to `sort=order` and `order:asc` (the `order` field might be a better name eventually).

It would be nice to have the opinion also of @KonstantinaStoikou and @kprzerwa 
I think that we should have a custom manipulation of this vocab, when pushing it to Elasticsearch, and put as first this

kprzerwa: for short term solution better would be option 1)
but since it is not the single case that we need a specific order in the dictionary, normally I would be in favour of option 2, however because we are supposed to integrate invenio-vocabularies later, the temporary solution seems more convenient for now.
Maybe also opening an issue (or RFC) in invenio-vocabularies would be helpful for the future integration

src/lib/forms/components/LanguageField.js
```diff
@@ -4,14 +4,15 @@ import { VocabularyField } from '@forms/core/VocabularyField';
...
-    const { multiple, type, fieldPath } = this.props;
+    const { multiple, type, fieldPath, required } = this.props;
     return (
       <VocabularyField
         multiple={multiple}
         type={type}
         fieldPath={fieldPath}
         label={multiple ? 'Languages' : 'Language'}
         placeholder={multiple ? 'Select languages...' : 'Select language'}
+        required={required}
```
kprzerwa, line 15: I have doubts about the languages being required. In case of a library having only books in one language, f.e. english it is unnecessary @ntarocco

src/lib/config/defaultConfig.js
```diff
@@ -241,6 +241,7 @@ export const RECORDS_CONFIG = {
       fields: {},
     },
     frontsiteMaxLinks: 5, // maximum number of links to show on details page
+    defaultLanguages: ['ENG'],
```
ntarocco, line 244: I see that language is used also in the series... I would go for a more global config, as for the default currency, WDYT?

jrcastro2: I agree, it makes sense to add it also to the series. Thanks!

src/lib/config/common.js
```diff
@@ -2,6 +2,8 @@
 
 export const DEFAULT_CURRENCY = 'CHF';
 
+export const DEFAULT_LANGUAGES = ['ENG'];
```
ntarocco, line 5: I would go for singular, because it is very misleading otherwise... WDYT?
```suggestion
export const DEFAULT_LANGUAGE = 'ENG';
```

kprzerwa: yes also we might make a mistake and assign a string instead of the list

src/lib/config/defaultConfig.js
```diff
@@ -30,6 +31,7 @@ export const APP_CONFIG = {
   REST_ENDOINTS_BASE_URL: process.env.REACT_APP_INVENIO_REST_ENDPOINTS_BASE_URL,
   SUCCESS_AUTO_DISMISS_SECONDS: 10,
   DEFAULT_CURRENCY: DEFAULT_CURRENCY,
+  DEFAULT_LANGUAGES: [DEFAULT_LANGUAGE],
```
kprzerwa, line 34: I think it should be string here, not an array

jrcastro2: I thought maybe someone would like to have more than one default language, but I see your point, I will change it. Thanks!

kprzerwa (approved):

PR #375, opened 2021-02-25
fixes: responsive

jrcastro2:
* Fixes minor responsive issues
* Fixes breadcrumbs alignment
* closes https://github.com/CERNDocumentServer/cds-ils/issues/24

### DIFF
--- a/src/lib/components/ILSMenu/ILSMenu.js
+++ b/src/lib/components/ILSMenu/ILSMenu.js
@@ -45,12 +45,12 @@ class ILSMenu extends Component {
 
     return (
       <>
-        <Media greaterThanOrEqual="tablet">
+        <Media greaterThanOrEqual="computer">
           <Dropdown item text={userMenuText} icon="caret down">
             {dropdownEntries}
           </Dropdown>
         </Media>
-        <Media at="mobile">
+        <Media lessThan="computer">
           <Dropdown item text={userMenuText} icon="bars">
             {dropdownEntries}
           </Dropdown>
@@ -78,7 +78,7 @@ class ILSMenu extends Component {
     return (
       <Overridable id="ILSMenu.layout" {...this.props}>
         <>
-          <Media greaterThanOrEqual="tablet">
+          <Media greaterThanOrEqual="computer">
             <Menu
               stackable
               borderless
@@ -123,7 +123,7 @@ class ILSMenu extends Component {
               </Container>
             </Menu>
           </Media>
-          <Media at="mobile">
+          <Media lessThan="computer">
             <Menu
               borderless
               inverted
--- a/src/lib/components/SearchBar/SearchBarILS.js
+++ b/src/lib/components/SearchBar/SearchBarILS.js
@@ -1,6 +1,8 @@
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 import { Input } from 'semantic-ui-react';
+import { screenIsWiderThan } from '@components/utils';
+import { Breakpoints } from '@components/Media';
 
 export class SearchBarILS extends Component {
   state = { currentValue: '' };
@@ -35,9 +37,11 @@ export class SearchBarILS extends Component {
       onPasteHandler,
       onChangeHandler,
       placeholder,
+      responsiveAutofocus,
       ...rest
     } = this.props;
     const { currentValue } = this.state;
+    const autofocus = screenIsWiderThan(Breakpoints.computer);
     return (
       <Input
         action={{
@@ -55,7 +59,11 @@ export class SearchBarILS extends Component {
         placeholder={placeholder}
         className={`${parentClass} ils-searchbar`}
         ref={(input) => {
-          this.focusInput = input;
+          if (responsiveAutofocus && autofocus) {
+            this.focusInput = input;
+          } else if (!responsiveAutofocus) {
+            this.focusInput = input;
+          }
         }}
         {...rest}
       />
@@ -70,6 +78,7 @@ SearchBarILS.propTypes = {
   onChangeHandler: PropTypes.func,
   placeholder: PropTypes.string,
   className: PropTypes.string,
+  responsiveAutofocus: PropTypes.bool,
 };
 
 SearchBarILS.defaultProps = {
@@ -78,4 +87,5 @@ SearchBarILS.defaultProps = {
   onChangeHandler: null,
   placeholder: '',
   className: '',
+  responsiveAutofocus: false,
 };
--- a/src/lib/components/utils.js
+++ b/src/lib/components/utils.js
@@ -1,3 +1,7 @@
 export const prettyPrintBooleanValue = (value) => {
   return value ? 'Yes' : 'No';
 };
+
+export const screenIsWiderThan = (pixels) => {
+  return window.matchMedia(`(max-width: ${pixels}px)`).matches ? false : true;
+};
--- a/src/lib/modules/Series/SeriesLiteratureSearch/SeriesLiteratureSearch.js
+++ b/src/lib/modules/Series/SeriesLiteratureSearch/SeriesLiteratureSearch.js
@@ -24,6 +24,7 @@ import {
   SearchBar,
 } from 'react-searchkit';
 import { Container, Divider, Loader } from 'semantic-ui-react';
+import { SearchBarOverridesMap } from '@components/SearchBar/SearchBarOverrides';
 import { qsBuilderForSeries } from './RequestSerializer';
 import { SeriesLiteratureSearchMobile } from './SeriesLiteratureSearchMobile';
 
@@ -60,6 +61,7 @@ export class SeriesLiteratureSearch extends React.Component {
           value={{
             ...SearchControlsOverridesMap,
             ...LiteratureSearchOverridesMap,
+            ...SearchBarOverridesMap,
           }}
         >
           <ReactSearchKit
@@ -74,6 +76,7 @@ export class SeriesLiteratureSearch extends React.Component {
                 <SearchBar
                   placeholder="Search for volumes or issues..."
                   {...invenioConfig.APP.SEARCH_BAR_PROPS}
+                  responsiveAutofocus
                 />
               </Container>
               <Media greaterThanOrEqual="computer">
--- a/src/lib/pages/frontsite/Documents/DocumentDetails/DocumentDetails.js
+++ b/src/lib/pages/frontsite/Documents/DocumentDetails/DocumentDetails.js
@@ -193,6 +193,7 @@ class DocumentDetails extends Component {
                 onSearchHandler={this.onSearchClick}
                 placeholder={invenioConfig.APP.HOME_SEARCH_BAR_PLACEHOLDER}
                 className="fs-headline"
+                responsiveAutofocus
               />
             </Container>
           </Container>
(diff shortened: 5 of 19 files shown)


src/lib/config/defaultConfig.js
```diff
@@ -549,6 +549,7 @@ export const RECORDS_CONFIG = {
       { value: 'ONE_WEEK', text: '1 week' },
       { value: 'TWO_WEEKS', text: '2 weeks' },
       { value: 'THREE_WEEKS', text: '3 weeks' },
+      { value: 'FOUR_WEEKS', text: '4 weeks' },
```
ntarocco, line 552: to be removed as discussed

src/lib/components/SearchBar/SearchBarILS.js
```diff
@@ -6,7 +6,8 @@ export class SearchBarILS extends Component {
   state = { currentValue: '' };
 
   componentDidMount() {
-    if (this.focusInput) {
+    const { autoFocus } = this.props;
+    if (autoFocus && this.focusInput) {
```
ntarocco, line 10: I guess you did it, but did we use the `focusInput.focus()` anywhere else and maybe needs fix?

src/lib/pages/frontsite/Home/Headline/HomeSearchBar.js
```diff
@@ -12,12 +13,25 @@ export class HomeSearchBar extends Component {
...
   render() {
     return (
-      <SearchBarILS
-        onPasteHandler={() => {}}
-        onSearchHandler={this.onSearchExecute}
-        placeholder={invenioConfig.APP.HOME_SEARCH_BAR_PLACEHOLDER}
-        className="fs-headline"
-      />
+      <>
+        <Media greaterThanOrEqual="computer">
```
ntarocco, line 17: it looks the same code to me, can we avoid the duplication?

src/lib/pages/frontsite/Literature/LiteratureSearch/LiteratureSearch.js
```diff
@@ -73,10 +73,23 @@ class LiteratureSearch extends Component {
             <>
               <Container fluid className="literature-search-container">
                 <Container>
-                  <SearchBar
-                    placeholder={invenioConfig.APP.HOME_SEARCH_BAR_PLACEHOLDER}
-                    {...invenioConfig.APP.SEARCH_BAR_PROPS}
-                  />
+                  <Media greaterThanOrEqual="computer">
```
ntarocco, line 76: same here

src/lib/pages/frontsite/Series/SeriesDetails/SeriesDetails.js
```diff
@@ -157,12 +157,23 @@ class SeriesDetails extends React.Component {
             executeSearch={this.onSearchClick}
           >
             <Container>
-              <SearchBarILS
-                onPasteHandler={() => {}}
-                onSearchHandler={this.onSearchClick}
-                placeholder={invenioConfig.APP.HOME_SEARCH_BAR_PLACEHOLDER}
-                className="fs-headline"
-              />
+              <Media greaterThanOrEqual="computer">
```
ntarocco, line 160: same here

src/semantic-ui/site/globals/site.overrides
```diff
@@ -142,7 +142,9 @@ a {
     margin-left: 1.6em;
   }
   &.frontsite-mobile {
+    position: absolute;
```
ntarocco, line 145: what is this solving? I am not a fan of negative margins! :)

src/lib/pages/frontsite/Documents/DocumentDetails/DocumentDetails.js
```diff
@@ -188,12 +188,23 @@ class DocumentDetails extends Component {
...
           <Container fluid className="literature-search-container">
             <Container>
-              <SearchBarILS
-                onPasteHandler={() => {}}
-                onSearchHandler={this.onSearchClick}
-                placeholder={invenioConfig.APP.HOME_SEARCH_BAR_PLACEHOLDER}
-                className="fs-headline"
-              />
+              <Media greaterThanOrEqual="computer">
+                <SearchBarILS
```
ntarocco, line 192: same here

src/lib/modules/Series/SeriesLiteratureSearch/SeriesLiteratureSearch.js
```diff
@@ -70,13 +70,13 @@ export class SeriesLiteratureSearch extends React.Component {
...
           >
             <>
-              <Container className="series-details-search-container">
-                <SearchBar
-                  placeholder="Search for volumes or issues..."
-                  {...invenioConfig.APP.SEARCH_BAR_PROPS}
-                />
-              </Container>
               <Media greaterThanOrEqual="computer">
+                <Container className="series-details-search-container">
```
ntarocco, line 74: same here, see comment below

src/lib/components/SearchBar/SearchBarILS.js
jrcastro2: I checked and it's only used in RSK but I already added the `autofocus={false}` in this cases

src/semantic-ui/site/globals/site.overrides
jrcastro2: It's only affecting the `environmentLabel` and it basically prevents it from interfiering with other components/tags (since it's not part of the production environment)
image

src/lib/components/ILSMenu/ILSMenu.js
```diff
@@ -45,12 +45,12 @@ class ILSMenu extends Component {
 
     return (
       <>
-        <Media greaterThanOrEqual="tablet">
+        <Media greaterThanOrEqual="computer">
```
kprzerwa, line 48: how does it look on the tablet after this change ?

src/lib/components/SearchBar/SearchBarILS.js
kprzerwa: does it still autofocus on desktop after this change ?

src/lib/pages/frontsite/Documents/DocumentDetails/DocumentDetails.js
```diff
@@ -180,6 +181,7 @@ class DocumentDetails extends Component {
     if (hasError && _get(error, 'response.status') === 404) {
       return <NotFound />;
     }
+    const autofocus = screenIsWiderThan(Breakpoints.computer);
```
kprzerwa, line 184: you could add this inside of SearchBarILS, for the component to recognise when it is suppose to use autofocus, not in every parent component that uses it. It would reduce the amount of code.

src/semantic-ui/site/globals/site.overrides
kprzerwa: maybe only for mobile we could display it like this, over the site title element (CERN Library Catalogue) instead of going with specific css (just an idea, if it is not worth the time, don't change it ;) )
Screenshot 2021-02-26 at 14 36 57

src/lib/components/ILSMenu/ILSMenu.js
jrcastro2: On an iPad (768x1024) it looks like this :
image

src/lib/pages/frontsite/Documents/DocumentDetails/DocumentDetails.js
jrcastro2: Yes, but since we only want this on the frontsite I thought it would be "cleaner" to not force this behaviour in the backoffice. Anyway if backoffice is not supossed to be responsive I guess we can move it inside SearchBarILS.

src/semantic-ui/site/globals/site.overrides
jrcastro2: We would still need specific css, since it gets cut on top:
image

src/lib/pages/frontsite/Documents/DocumentDetails/DocumentDetails.js
kprzerwa: the backoffice is not supposed to be used with responsive. you can still add it as an optional prop to SearchBar and SearchBarILS to have the control to turn off/on the focus, but  IMO the check if the screen is wide or not should be only done in the SeachBar, not the parent components (because it's not their purpose to know about details of displaying the SearchBar)

jrcastro2: Yes, you are absolutely right about your point, I will change it. Thanks!

src/lib/components/utils.js
```diff
@@ -1,3 +1,7 @@
 export const prettyPrintBooleanValue = (value) => {
   return value ? 'Yes' : 'No';
 };
+
+export const screenIsWiderThan = (pixeles) => {
```
ntarocco, line 5: Good solution in the end!

Very minor typo, but you can keep it as it is or fix in another PR!
```suggestion
export const screenIsWiderThan = (pixels) => {
```

src/semantic-ui/site/collections/breadcrumb.overrides
```diff
@@ -1,3 +1,8 @@
 /*******************************
  Breadcrumb Overrides - REACT-INVENIO-APP-ILS
 *******************************/
+
+
+.ui.breadcrumb {
+    line-height: 1.1;
```
ntarocco, line 7: Thanks!

ntarocco (approved):

src/lib/components/SearchBar/overriden/ResponsiveSearchBar.js
```diff
@@ -0,0 +1,176 @@
...
+      executeSearch,
+      onBtnSearchClick,
+      onInputChange,
+      onKeyPress,
+      overridableId,
+      placeholder,
+      uiProps,
+    } = this.props;
+    let shouldFocus = autofocus;
+    if (responsiveAutofocus) {
```
jrcastro2, line 50: Note for reviewers: This is the same searchBar from RSK, the only difference is the new `responsiveAutofocus` prop.

src/lib/components/SearchBar/overriden/ResponsiveSearchBar.js
```diff
@@ -0,0 +1,176 @@
...
+class SearchBar extends Component {
+  constructor(props) {
+    super(props);
+    this.updateQueryString = this.props.updateQueryString;
+    this.state = {
+      currentValue: this.props.queryString || '',
+    };
+  }
+
+  onInputChange = (queryString) => {
```
kprzerwa, line 16: are these methods needed? aren't they implemented in RSK?

jrcastro2: Hmm, as I understand it, we override the entire RSK [searchBar](https://github.com/inveniosoftware/react-searchkit/blob/master/src/lib/components/SearchBar/SearchBar.js#L155), so we need all the methods? or not?

jrcastro2: Following @ntarocco suggestion I parameterized the RSK SearchBar and it works correclty. Thanks!

ntarocco (approved):

PR #380, opened 2021-02-25
Backoffice dashboard: add pending literature requests

### LINKED ISSUE #378: backoffice dashboard: add new panel for pending "New literature requests"

ntarocco:
* adds a new dashboard panel to display the pending new
  literature requests
* refactors all dashboard components to remove redux
* closes #378


ntarocco:
I have removed the tests. If you think they were useful I can put them back.

src/lib/pages/backoffice/Home/lists/IdleLoansList.js
```diff
@@ -1,31 +1,61 @@
...
 
-export default class IdleLoansList extends Component {
+export class IdleLoansList extends Component {
+  constructor(props) {
+    super(props);
+
+    this.query = loanApi
+      .query()
+      .withState(invenioConfig.CIRCULATION.loanRequestStates)
+      .withUpdated({
```
ntarocco, line 24: @kprzerwa I am not sure about this, I would like you to review it. I copied from the existing, did not change anything.
The panel here is for `Loan requests pending since more than 10 days`: shouldn't we check `request_start_date` less than 10 days instead of the updated?

kprzerwa: it was checking `updated` on purpose, to see if there are some idle requests (as in overlooked and not changed/processed by librarians)

src/lib/pages/backoffice/Home/Home.js
```diff
@@ -31,14 +32,22 @@ export default class Home extends Component {
 
         <Grid.Row columns={2}>
           <Grid.Column>
+            <Divider hidden />
```
kprzerwa, line 35: thanks! does it look better with the white spaces ?

kprzerwa (approved):

src/lib/pages/backoffice/Home/Home.js
ntarocco: Not much really :( I think the fact that the boxes do not have the same size is bothering me.
I would prefer to have boxes of a fixed size, even if empty, and not having them jumping around depending on how many rows are there...

src/lib/pages/backoffice/Home/lists/IdleLoansList.js
ntarocco: I am not sure I understand :( It could be they are updated for other reasons, I am not sure the updated date is reliable.
Why not using the request_start_date?

kprzerwa: indeed it could be used instead. After some thinking, I remember that we added the `request_start_date` later than we created the dashboard so we used initially `updated` for this feature

PR #381, opened 2021-02-26
replace locations forms with RJSF

ntarocco:
* replaces location and internal location forms
* adds a disable state to the form when submitting


ntarocco:
Internal Location
Screenshot 2021-02-26 at 19 51 53

Location
Screenshot 2021-02-26 at 19 46 19
Screenshot 2021-02-26 at 19 46 26

src/lib/forms/rjsf/fields/RJSFLabelField.js
```diff
@@ -0,0 +1,35 @@
+import HiddenWidget from '@rjsf/core/dist/cjs/components/widgets/HiddenWidget';
+import _capitalize from 'lodash/capitalize';
+import PropTypes from 'prop-types';
+import React, { Component } from 'react';
+import { Header } from 'semantic-ui-react';
+
+/**
+ * Custom field to display a field as a label and the value in the hidden field
```
kprzerwa, line 8: I am not sure if I understand how this field should behave, maybe we could see it in a call

kprzerwa (approved):

src/lib/forms/rjsf/widgets/RJSFTimeWidget.js
```diff
@@ -0,0 +1,63 @@
...
+      readonly,
+      label,
+      onChange,
+    } = this.props;
+    return (
+      <Form.Field
+        id={id}
+        key={id}
+        required={required}
+        disabled={disabled || readonly}
```
jrcastro2, line 26: Question: `readonly` is not used anywhere if I am not mistaken. Do we have it to improve readability in the future?

jrcastro2 (approved):

src/lib/forms/rjsf/widgets/RJSFTimeWidget.js
ntarocco: I am not sure what you mean exactly with `improve readability`, but this is part of the API contract of the RJSF widgets APIs, so if we use it in a schema, it is already working correctly.

src/lib/forms/rjsf/fields/RJSFLabelField.js
ntarocco: Yes, in a couple of words, if you have a readonly input the text is grayed out, almost unreadable.
The idea is to display it a label (see the name of the days in the screenshot Monday, ...) but it has to be part of the form and sent to the backend so it adds an hidden field with as value the name of the day.

PR #392, opened 2021-03-05
[DEPENDS on IAILS#1058] literatureRequest: form

jrcastro2:
* Adds publisher field
* Rearranges fields
* closes https://github.com/CERNDocumentServer/cds-ils/issues/355
* needs https://github.com/inveniosoftware/invenio-app-ils/pull/1058
image


src/lib/forms/core/StringField.js
```diff
@@ -15,13 +16,32 @@ export class StringField extends Component {
...
     } = props;
-
+    const { required } = uiProps;
+    let popupLabelClassName = 'display-flex';
+    if (popupHelper && required) {
+      popupLabelClassName += ' required field';
+    }
     return (
       <Form.Field inline={inline} width={width} disabled={isSubmitting}>
+        {popupHelper ? (
```
ntarocco, line 38: We are going to specific here IMHO: some fields have a popup, some other a label.
We are abusing of generic components and hardcoding a specific layout of a specific form. At this point, it is better to created a specific `DocReqStringField` that wraps the original/generic one and adds functionality.

A better option is probably to simply pass a `component` prop, and you pass the entire label or popup from the caller.

Remember that the API interface of your cmp/function/method has to be the purest possible and you have to maximise re-usability. The inner implementation should know nothing of the caller or in which context it will be used.

src/semantic-ui/site/globals/site.overrides
```diff
@@ -215,6 +215,24 @@ a {
   }
 }
 
+.custom-label {
```
ntarocco, line 218: I fear that this is too generic, while it is used in a very specific context.
Is this re-usable anywhere else? If not:
```
.<doc req form class> {
    .custom-label
}
```
or even better prefer ids `#` whenever possible, the rendering is faster.

If it is re-usable, then I would call it with some more meaningful names.

src/lib/pages/frontsite/DocumentRequests/DocumentRequestForm.js
```diff
@@ -203,8 +245,8 @@ class DocumentRequestForm extends Component {
           <>
             <Header as="h1">Request new literature</Header>
             <p>
-              Fill in the form below to request a new literature from the
-              library.
+              Fill in the form below to request to loan, purchase or suggest a
```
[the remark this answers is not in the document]
jrcastro2: Yes, I agree about the message, instead of adding it to the config we could add a prop for the `title` and override it with the `parametrize` feature of `react-override`, same  for the `labelSubtitle`, as I did for the `infoMessage` in the [`requestType`](https://github.com/CERNDocumentServer/cds-ils/pull/368/files#diff-562f0ff740e031c21e40f9514b78f33498423e3936cd3655f494226b5001302dR15-R36), WDYT?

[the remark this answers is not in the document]
kprzerwa: I think I would go with overriding rather than adding it to the config. It would be more flexible

src/lib/forms/core/SelectField.js
```diff
@@ -127,6 +129,7 @@ SelectField.propTypes = {
   options: PropTypes.array.isRequired,
   label: PropTypes.string,
   width: PropTypes.number,
+  customLabel: PropTypes.object,
```
ntarocco, line 132: What about:
```suggestion
  customLabel: PropTypes.PropTypes.oneOfType([
    PropTypes.string,
    PropTypes.node,,
    ...
```

ntarocco (approved):

kprzerwa (commented):
One last comment:
the highlighted fields should be more on top, since they are more important than journal related fields
110339304-50f1c080-8028-11eb-8cd7-b47525563c70

PR #403, opened 2021-03-12
[Merge with IAILS#1064] Replace Document form with RJSF and various fixes

### LINKED ISSUE #303: Patron details page - backoffice - edits
Add the document title column in the ongoing loans table in the Backoffice Patron view

Pasted_File

### LINKED ISSUE #362: Document editor: make extension field of type list editable
As discovered testing extension fields of type `list`, they are not editable. In the editor they are rendered as a normal input field for one string.

See [this comment](https://github.com/inveniosoftware/invenio-app-ils/pull/1032#issuecomment-779330588) for more info.

It will probably come for free if the document editor is moved to React-JSONSchema-Forms but it is not an easy task.

ntarocco:
* replaces the Document form with RJSF
* fixes creation/edition of extension fields
* harmonizes backoffice tables headers/titles to have similar names
* adds support for ES 6 / ES 7
* harmonizes how boolean values are displayed
* removes dead code
* removes yup dependency
* closes #303
* closes #362


ntarocco:
In case of too many authors:

Screenshot 2021-03-12 at 18 14 57

KonstantinaStoikou (approved):
LGTM, the only comment I have is that in some cases e.g.:
Screenshot 2021-03-15 at 16 06 56
the buttons `Add Item` are very close to each other and it is not very clear to which field they are referring to. Is it possible to change the text to something like `Add author`? Or maybe to move the button next to the title of the field?

ntarocco:
> LGTM, the only comment I have is that in some cases e.g.:
> Screenshot 2021-03-15 at 16 06 56
> the buttons `Add Item` are very close to each other and it is not very clear to which field they are referring to. Is it possible to change the text to something like `Add author`? Or maybe to move the button next to the title of the field?

wow, thank you for checking it out and checking!!! 👍🏼 
Unfortunately not much possible. I was hoping that with the release of RJSF 2.5.0  we could finall override the `Array` template (see my PR [here](https://github.com/rjsf-team/react-jsonschema-form/pull/2231)) but they did a mistake publishing on NPM and we cannot override yet. As soon as they do, we can override the template probably and do something...

I think the best we can not now is to avoid to have array fields as last field inside another array field, but to have a text for example instead.

src/lib/api/utils.js
```diff
@@ -146,3 +146,12 @@ export const searchReady = async () => {
...
   );
 };
+
+/**
+ * Get the total number of results from a search API call, to
+ * support ES 6 and ES 7.
+ * @param {*} responseData the response data object
+ * @returns the total number of results
+ */
+export const getSearchTotal = (hits) =>
```
kprzerwa, line 156: in the long term as a solution, shouldn't this be the responsibility of the backend to normalise it? f.e. invenio-search, to have the consistent structure on REST, not depending on the ES version? Should frontend be avare of ES version?

kprzerwa (approved):
wow so many deleted files :)

src/lib/api/utils.js
ntarocco: Very valid point! You are totally right!

PR #421, opened 2021-03-24
display legacy ids in backoffice details pages

ntarocco:
* displays legacy id field when set
* upgrades RJSF to latest and enables ArrayFieldTemplate override
* fixes to providers cmps


ntarocco:
Tested details pages, searches and that when editing via form the legacy id is kept.

kprzerwa (approved):

Issue #223, opened 2020-10-06
Add `login_required` field and display e-items links


ntarocco:
2 tasks:

* in the backend, a new boolean field `login_required` for each e-item URL is added. Add it to the form and also display in the backoffice e-item details page (see [here](https://github.com/inveniosoftware/invenio-app-ils/pull/947))
* The e-items links should be displayed in the frontsite document details page. It looks like that only `None` is displayed now, see [here](https://github.com/inveniosoftware/react-invenio-app-ils/blob/ed6c9ea00ea3964ebd6a609575a56646949efd4f/src/lib/modules/Document/DocumentLinks.js#L28). The cmp should be fixed following the rules below.

### Open access

Files are always visible and clickable.

### Closed access

To see and download files, you have to login.

### Open access

URLs are always visible and clickable.

### Closed access

URLs are always visible and clickable.

PR #426, opened 2021-04-06
[BLOCKED]series: add login_required to access_urls

KonstantinaStoikou:
Closes https://github.com/CERNDocumentServer/cds-migrator-kit/issues/129
Depends on https://github.com/inveniosoftware/invenio-app-ils/pull/1077

Screenshot 2021-04-06 at 15 07 43
Screenshot 2021-04-06 at 15 47 08

### DIFF
--- a/src/lib/modules/Series/SeriesLinks.js
+++ b/src/lib/modules/Series/SeriesLinks.js
@@ -13,7 +13,7 @@ export const SeriesLinks = ({ accessUrls, urls }) => {
       <Divider horizontal>Access online</Divider>
       <LiteratureAccessUrls urls={accessUrls} />
       <Divider horizontal>Links</Divider>
-      <SeriesUrls url={urls} />
+      <SeriesUrls urls={urls} />
     </>
   );
 };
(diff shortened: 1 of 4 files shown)


src/lib/modules/Document/DocumentEItemUrls.js
```diff
@@ -19,22 +19,29 @@ export class DocumentEItemUrls extends Component {
     );
   };
 
-  userCanSeeFiles = (eitem) => {
+  userCanSeeEItems = (eitem) => {
```
ntarocco, line 22: I am not sure about this change and below.
From what I remember, it was that when `open_access: False`, links are still visible but not files.
Can you check the issue when this was implemented?

KonstantinaStoikou: You are right. [Here](https://github.com/inveniosoftware/react-invenio-app-ils/issues/223) is the issue. For the `access_urls` of series, when `open_access` is False I am not displaying the urls for not logged in users. Is it right?

kprzerwa (approved):

PR #445, opened 2021-05-17
Clear checkin bar

kprzerwa:
for reviewers: check only first commit https://github.com/inveniosoftware/react-invenio-app-ils/commit/12183bf9101cf9b8bf5b70693dc54cdaea3472f9
closes https://github.com/CERNDocumentServer/cds-ils/issues/486


src/lib/components/backoffice/DeleteRecordModal/DeleteRecordModal.js
```diff
@@ -60,7 +60,7 @@ export default class DeleteRecordModal extends Component {
         key={hit.id}
         as="a"
         onClick={() => onRefClick(hit.id)}
-        target="blank"
+        target="_blank"
```
ntarocco, line 63: 😱

src/lib/modules/Series/SeriesCard.js
```diff
@@ -42,12 +42,10 @@ export class SeriesCard extends Component {
       : null;
     return (
       <Card
-        link
         centered
         className="fs-book-card"
-        onClick={() =>
-          goTo(FrontSiteRoutes.seriesDetailsFor(data.metadata.pid))
-        }
+        as={Link}
```
ntarocco, line 47: so apparently this fixes the opening in new tab?

src/lib/pages/backoffice/Actions/CheckIn/ItemsSearch/ItemsSearch.js
```diff
@@ -20,6 +25,7 @@ export default class ItemsSearch extends Component {
         <SearchBarILS
           onSearchHandler={this.executeCheckinAndClearInput}
           placeholder="Scan physical copy barcode to check-in..."
+          ref={this.searchBarRef}
```
ntarocco, line 28: Ideally, this is not the recommended way in React. Ideally, the outer cmp should have a state that is injected in the inner cmps. Outer components should not know internal implementations of other cmps. If it is not exposed via a prop, then it is a private method.

Having said so, I understand it is tricky to refactor it now.
Please keep this in mind because it should not become a pattern and my fear is that we copy/paste this and it becomes an habit.

ntarocco (approved):

src/lib/modules/Series/SeriesCard.js
kprzerwa: this was only the old code leftover, the link way is cleaner, the actual fix is in the `DocumentCard`

src/lib/pages/backoffice/Actions/CheckIn/ItemsSearch/ItemsSearch.js
kprzerwa: that is true, it is not the best way to deal with this problem, but I have chosen it because of the alternative being tricky to refactor, especially for such a small given value

PR #459, opened 2021-07-07
CheckIn

Pineirin:
Add autofocus to the input form. Invert order of the returned loans list and highlight the location of each book in that list. Visual ui changes.
closes https://github.com/CERNDocumentServer/cds-ils/issues/531
Screenshot from 2021-08-03 13-08-37

### DIFF
--- a/src/lib/components/SearchBar/SearchBarILS.js
+++ b/src/lib/components/SearchBar/SearchBarILS.js
@@ -1,15 +1,21 @@
 import PropTypes from 'prop-types';
-import React, { Component } from 'react';
+import React, { Component, createRef } from 'react';
 import { Input } from 'semantic-ui-react';
 import { screenIsWiderThan } from '@components/utils';
 import { Breakpoints } from '@components/Media';
 
 export class SearchBarILS extends Component {
+  inputRef = createRef();
+
   state = { currentValue: '' };
 
   componentDidMount() {
-    if (this.focusInput) {
-      this.focusInput.focus();
+    const { responsiveAutofocus } = this.props;
+    const autofocus = screenIsWiderThan(Breakpoints.computer);
+    if (responsiveAutofocus && autofocus) {
+      this.inputRef.current.focus();
+    } else if (!responsiveAutofocus) {
+      this.inputRef.current.focus();
     }
   }
 
@@ -43,11 +49,11 @@ export class SearchBarILS extends Component {
       onPasteHandler,
       onChangeHandler,
       placeholder,
+      ref,
       responsiveAutofocus,
       ...rest
     } = this.props;
     const { currentValue } = this.state;
-    const autofocus = screenIsWiderThan(Breakpoints.computer);
     return (
       <Input
         action={{
@@ -65,13 +71,7 @@ export class SearchBarILS extends Component {
         size="big"
         placeholder={placeholder}
         className={`${parentClass} ils-searchbar`}
-        ref={(input) => {
-          if (responsiveAutofocus && autofocus) {
-            this.focusInput = input;
-          } else if (!responsiveAutofocus) {
-            this.focusInput = input;
-          }
-        }}
+        ref={this.inputRef}
         {...rest}
       />
     );
--- a/src/lib/modules/Items/backoffice/ItemListEntry.js
+++ b/src/lib/modules/Items/backoffice/ItemListEntry.js
@@ -6,7 +6,7 @@ import _isEmpty from 'lodash/isEmpty';
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 import { Link } from 'react-router-dom';
-import { Button, Grid, Header, Item, List } from 'semantic-ui-react';
+import { Button, Grid, Header, Item, List, Label } from 'semantic-ui-react';
 
 class ItemCirculation extends Component {
   render() {
@@ -61,6 +61,7 @@ export class ItemListEntry extends Component {
       record: item,
       withPendingLoans,
       showPreviousLoan,
+      highlightLocation,
       target,
     } = this.props;
     const classes = withPendingLoans ? 'bkg-yellow' : '';
@@ -85,16 +86,36 @@ export class ItemListEntry extends Component {
                 </Header>
                 <List>
                   <List.Item>
-                    <List.Content>
-                      {item.metadata.internal_location.location.name} -{' '}
-                      {item.metadata.internal_location.name}
-                    </List.Content>
+                    {highlightLocation ? (
+                      <List.Content>
+                        <Label color="blue" tag>
+                          <strong>
+                            {item.metadata.internal_location.location.name} -{' '}
+                            {item.metadata.internal_location.name}
+                          </strong>
+                        </Label>
+                      </List.Content>
+                    ) : (
+                      <List.Content>
+                        {item.metadata.internal_location.location.name} -{' '}
+                        {item.metadata.internal_location.name}
+                      </List.Content>
+                    )}
                   </List.Item>
                   <List.Item>
-                    <List.Content>
-                      {' '}
-                      shelf <strong>{item.metadata.shelf}</strong>
-                    </List.Content>
+                    {highlightLocation ? (
+                      <List.Content>
+                        {' '}
+                        <Label color="blue" tag>
+                          <strong>shelf {item.metadata.shelf}</strong>
+                        </Label>
+                      </List.Content>
+                    ) : (
+                      <List.Content>
+                        {' '}
+                        shelf <strong>{item.metadata.shelf}</strong>
+                      </List.Content>
+                    )}
                   </List.Item>
                 </List>
               </Item.Meta>
@@ -151,9 +172,9 @@ export class ItemListEntry extends Component {
               {withPendingLoans && (
                 <Item.Extra>
                   <Button
+                    primary
                     compact
                     floated="right"
-                    color="blue"
                     as={Link}
                     target="_blank"
                     to={BackOfficeRoutes.documentDetailsFor(
@@ -167,7 +188,11 @@ export class ItemListEntry extends Component {
             </Grid.Column>
           </Grid>
         </Item.Content>
-        <div className="pid-field">#{item.metadata.pid}</div>
+        {highlightLocation ? (
+          <div className="pid-field pid-field-top">#{item.metadata.pid}</div>
+        ) : (
+          <div className="pid-field">#{item.metadata.pid}</div>
+        )}
       </Item>
     );
   }
@@ -177,11 +202,13 @@ ItemListEntry.propTypes = {
   record: PropTypes.object.isRequired,
   withPendingLoans: PropTypes.bool,
   showPreviousLoan: PropTypes.bool,
+  highlightLocation: PropTypes.bool,
   target: PropTypes.string,
 };
 
 ItemListEntry.defaultProps = {
   withPendingLoans: false,
   showPreviousLoan: false,
+  highlightLocation: false,
   target: '',
 };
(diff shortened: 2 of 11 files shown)


src/lib/pages/backoffice/Actions/CheckIn/CheckedInItems/CheckedInItems.js
```diff
@@ -12,7 +12,8 @@ export default class CheckedInItems extends Component {
 
   renderItems = () => {
     const { items } = this.props;
-    const entries = items.map((item) => {
+
+    const entries = items.reverse().map((item) => {
```
ntarocco, line 16: This is probably ok in most of the cases, but it might not be enough.
Here you are reversing the list client side, in place. The issue is that Elasticsearch will only give you the first `x` results. If the item to have as first is at the `x+1` position, the code above will not work because Elasticsearch did not return it.

It would be better to change the Elasticsearch query sort order instead, if possible.

src/semantic-ui/site/elements/list.overrides
```diff
@@ -32,6 +32,10 @@ List Overrides - REACT-INVENIO-APP-ILS
     padding: 0.5em;
   }
 
+  &.metadata-sharp{
```
ntarocco, line 35: To be discussed if bold+black is enough. In any case, instead of `sharp`, why not creating a class `.color-black` that can be eventually re-used?

src/lib/components/SearchBar/SearchBarILS.js
```diff
@@ -65,13 +55,6 @@ export class SearchBarILS extends Component {
         size="big"
         placeholder={placeholder}
         className={`${parentClass} ils-searchbar`}
-        ref={(input) => {
```
kprzerwa, line 68: are you sure we want to remove the focus from this component which is widely used across the whole application ?

src/lib/modules/Items/backoffice/ItemListEntry.js
```diff
@@ -75,25 +75,29 @@ export class ItemListEntry extends Component {
...
                     truncateLines={1}
                   />
                 </Header>
-                <List>
+                <List className="metadata-sharp">
                   <List.Item>
                     <List.Content>
-                      {item.metadata.internal_location.location.name} -{' '}
-                      {item.metadata.internal_location.name}
+                    <div class="ui label">
```
kprzerwa, line 89: in this way you are using the pure semantic ui, which is not what we want to do, since we have a library which provides the components for react. Please check react semantic ui library documentation for more details.

src/lib/modules/Items/backoffice/ItemListEntry.js
```diff
@@ -75,25 +75,29 @@ export class ItemListEntry extends Component {
...
+            <Grid.Column computer={5} largeScreen={5}>
               <Item.Meta className="metadata-fields">
                 <Header as="h5">
                   <LiteratureTitle
                     title={item.metadata.document.title}
                     truncateLines={1}
                   />
                 </Header>
-                <List>
+                <List className="metadata-sharp">
```
kprzerwa, line 86: do we still need this class, since below you use `<strong>` ?

src/lib/modules/Items/backoffice/ItemListEntry.js
```diff
@@ -75,25 +75,29 @@ export class ItemListEntry extends Component {
...
+                      <strong>{item.metadata.internal_location.location.name} -{' '}
+                      {item.metadata.internal_location.name}</strong>
+                    </div>
                     </List.Content>
                   </List.Item>
                   <List.Item>
                     <List.Content>
                       {' '}
-                      shelf <strong>{item.metadata.shelf}</strong>
+                      <div class="ui label">
```
kprzerwa, line 98: same as above

src/semantic-ui/site/elements/list.overrides
```diff
@@ -32,6 +32,12 @@ List Overrides - REACT-INVENIO-APP-ILS
     padding: 0.5em;
   }
 
+  &.metadata-sharp > * > * > *{
```
kprzerwa, line 35: in terms of CSS and semantic UI combination, the parent child selectors usage is quite dangerous - if someone changes the structure, or in the new version of semantic UI we will have more elements in this hierarchy, this stylesheet will not be applied correctly. is there any existing out of the box component you could use instead of the custom CSS? (custom CSS should be used only as a last resort, if there is no other solution)

src/lib/components/SearchBar/SearchBarILS.js
```diff
@@ -1,18 +1,10 @@
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 import { Input } from 'semantic-ui-react';
-import { screenIsWiderThan } from '@components/utils';
-import { Breakpoints } from '@components/Media';
 
 export class SearchBarILS extends Component {
   state = { currentValue: '' };
 
-  componentDidMount() {
```
kprzerwa, line 10: are you sure we want this removed ?

src/lib/components/SearchBar/SearchBarILS.js
```diff
@@ -43,11 +35,9 @@ export class SearchBarILS extends Component {
       onPasteHandler,
       onChangeHandler,
       placeholder,
-      responsiveAutofocus,
       ...rest
     } = this.props;
     const { currentValue } = this.state;
-    const autofocus = screenIsWiderThan(Breakpoints.computer);
```
kprzerwa, line 50: and this one ?

src/lib/modules/Items/backoffice/ItemListEntry.js
```diff
@@ -140,20 +150,12 @@ export class ItemListEntry extends Component {
...
-                to={BackOfficeRoutes.documentDetailsFor(
-                  item.metadata.document_pid
-                )}
-              >
-                <DocumentIcon /> Document
-              </Link>
               {withPendingLoans && (
                 <Item.Extra>
                   <Button
+                    primary
```
kprzerwa, line 156: 👍

src/lib/modules/Items/backoffice/ItemListEntry.js
```diff
@@ -78,22 +78,33 @@ export class ItemListEntry extends Component {
             <Grid.Column computer={6} largeScreen={6}>
               <Item.Meta className="metadata-fields">
                 <Header as="h5">
+                <Link
```
kprzerwa, line 81: how this influence the Items search view ? did you have any notification of changed snaphots in ItemSearch (in tests)?

src/semantic-ui/site/views/item.overrides
```diff
@@ -131,7 +125,7 @@ Item Overrides - REACT-INVENIO-APP-ILS
       }
 
       &.bkg-yellow {
-        background-color: #fff4bd !important;
+        background-color: ##ffffe6 !important;
```
kprzerwa, line 128: does it affect any other places in the application ?

Pineirin: No

src/lib/modules/Items/backoffice/ItemListEntry.js
Pineirin: I didn't, it seems to work fine, but I will double-check.

src/lib/modules/Items/backoffice/ItemListEntry.js
```diff
@@ -167,7 +204,11 @@ export class ItemListEntry extends Component {
             </Grid.Column>
           </Grid>
         </Item.Content>
-        <div className="pid-field">#{item.metadata.pid}</div>
+        {checkIn ? (
+          <div className="pid-field pid-field-top">#{item.metadata.pid}</div>
```
kprzerwa, line 208: actually it would be enough if you just removed all the classes here, without creating a new one

src/lib/modules/Items/backoffice/ItemListEntry.js
```diff
@@ -78,23 +79,57 @@ export class ItemListEntry extends Component {
             <Grid.Column computer={6} largeScreen={6}>
               <Item.Meta className="metadata-fields">
                 <Header as="h5">
-                  <LiteratureTitle
-                    title={item.metadata.document.title}
-                    truncateLines={1}
-                  />
+                  {checkIn ? (
```
kprzerwa, line 82: I think this change should be reverted, since it does not add so much in terms of the user experience, and it might be confusing to have this element in different places, depending on the location. I think we should keep things as much harmonised as possible

src/lib/modules/Items/backoffice/ItemListEntry.js
```diff
@@ -78,23 +79,57 @@ export class ItemListEntry extends Component {
...
+                    />
+                  )}
                 </Header>
                 <List>
                   <List.Item>
-                    <List.Content>
-                      {item.metadata.internal_location.location.name} -{' '}
-                      {item.metadata.internal_location.name}
-                    </List.Content>
+                    {checkIn ? (
```
kprzerwa, line 103: you are on a good track.
having in mind the previous comment, I would change `checkin` prop name to `highlightLocation` and use it only for this purpose. It might come up handy and more flexible if we want to highlight the location in the other places this component is used.

src/lib/modules/Items/backoffice/ItemListEntry.js
```diff
@@ -140,20 +175,22 @@ export class ItemListEntry extends Component {
...
             <Grid.Column computer={2} largeScreen={2} textAlign="right">
-              <Link
-                target="_blank"
-                to={BackOfficeRoutes.documentDetailsFor(
-                  item.metadata.document_pid
-                )}
-              >
-                <DocumentIcon /> Document
-              </Link>
+              {!checkIn && (
```
kprzerwa, line 178: I would revert this change, check the first comment for the explanation

src/lib/pages/backoffice/Actions/CheckIn/CheckedInItems/CheckedInItems.js
```diff
@@ -19,6 +19,7 @@ export default class CheckedInItems extends Component {
           key={item.id}
           withPendingLoans={item.has_pending_loans}
           showPreviousLoan
+          checkIn
```
kprzerwa, line 22: you didn't rename this one

src/lib/pages/frontsite/Documents/DocumentDetails/DocumentPanel/DocumentPanel.js
```diff
@@ -65,13 +65,17 @@ class DocumentPanel extends Component {
...
-                      {doc.metadata.imprint?.publisher
-                        &&
-                      <><div
-                        className={`default-margin-bottom`}>
-                        Published by <b>{doc.metadata.imprint?.publisher}</b>
-                        {doc.metadata.publication_year && (<b>{doc.metadata.publication_year}</b>)}
-                      </div></>}
+                      {doc.metadata.imprint?.publisher && (
+                        <>
+                          <div className={`default-margin-bottom`}>
```
kprzerwa, line 70:
```suggestion
                          <div className="default-margin-bottom">
```

src/lib/pages/frontsite/Documents/DocumentDetails/DocumentPanel/DocumentPanelMobile.js
```diff
@@ -50,13 +50,16 @@ class DocumentPanelMobile extends Component {
...
-                  {doc.metadata.imprint?.publisher
-                    &&
-                  <><div
-                    className={`default-margin-bottom`}>
-                    Published by <b>{doc.metadata.imprint?.publisher}</b>
-                    {doc.metadata.publication_year && (<b>{doc.metadata.publication_year}</b>)}
-                  </div></>}
+                  {doc.metadata.imprint?.publisher && (
+                    <>
+                      <div className={`default-margin-bottom`}>
```
kprzerwa, line 55:
```suggestion
                      <div className="default-margin-bottom">
```

PR #462, opened 2021-07-12
Show a warning for overbooked documents during check out

Pineirin:
Show a warning in the loans of overbooked documents.
closes:  https://github.com/CERNDocumentServer/cds-ils/issues/523
Loan of an overbooked document.
Screenshot from 2021-07-14 12-37-15
Loan of a non overbooked document.
Screenshot from 2021-07-14 12-36-45

### DIFF
--- a/src/lib/pages/backoffice/Loan/LoanDetails/LoanDetails.js
+++ b/src/lib/pages/backoffice/Loan/LoanDetails/LoanDetails.js
@@ -5,6 +5,7 @@ import { Loader } from '@components/Loader';
 import { Error } from '@components/Error';
 import { CurrentItem } from './CurrentItem';
 import { LoanHeader } from './LoanHeader';
+import { LoanOverbookedWarning } from './LoanOverbookedWarning';
 import { Loan } from './Loan';
 import { AvailableItems } from './AvailableItems';
 import { LoanActionMenu } from './LoanActionMenu';
@@ -54,6 +55,7 @@ export default class LoanDetails extends Component {
                     <Grid.Column width={13}>
                       <Container className="spaced">
                         <Container className="spaced">
+                          {data.metadata?.document_pid && <LoanOverbookedWarning documentPid={data.metadata.document_pid} />}
                           <Loan />
                           <CurrentItem />
                           <AvailableItems
--- a/src/lib/pages/backoffice/Loan/LoanDetails/LoanOverbookedWarning/LoanOverbookedWarning.js
+++ b/src/lib/pages/backoffice/Loan/LoanDetails/LoanOverbookedWarning/LoanOverbookedWarning.js
@@ -1,0 +1,68 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { withCancel } from '@api/utils';
+import { documentApi } from '@api/documents';
+import { Message } from 'semantic-ui-react';
+import { Link } from 'react-router-dom';
+import {
+  BackOfficeRoutes,
+} from '@routes/urls';
+
+const OVERBOOKED_MSG = (documentPid) => <>This literature is <Link to={BackOfficeRoutes.documentDetailsFor(documentPid)}>overbooked</Link></>
+const ERROR_MSG = "Cannot verify if this literature is overbooked. Please refresh the page."
+
+export default class LoanOverbookedWarning extends Component {
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      hasError: false,
+      msg: ''
+    };
+  }
+
+  componentDidMount() {
+    this.fetchCirculation();
+  }
+
+  componentWillUnmount() {
+    this.cancellableCirculation && this.cancellableCirculation.cancel();
+  }
+
+  fetchCirculation = async () => {
+    const { documentPid } = this.props;
+    try {
+      this.cancellableCirculation = withCancel(
+        documentApi.get(documentPid)
+      );
+      const response = await this.cancellableCirculation.promise;
+      const circulation = response.data.metadata.circulation;
+      if (circulation.overbooked){
+        this.setState({ hasError: false, msg: OVERBOOKED_MSG(documentPid)});
+      }
+    } catch (error) {
+      if (error !== 'UNMOUNTED') {
+        this.setState({ hasError: true, overbooked: OVERBOOKED_STATES.ERROR });
+      }
+    }
+  };
+
+
+  render() {
+    const { hasError, msg } = this.state;
+    const color = hasError ? 'red' : 'yellow';
+    return (
+      <>
+      {msg &&
+        <Message color={color}>
+          {msg}
+        </Message>
+      }
+      </>
+    );
+  }
+}
+
+LoanOverbookedWarning.propTypes = {
+  documentPid: PropTypes.string.isRequired,
+};
--- a/src/lib/pages/backoffice/Loan/LoanDetails/LoanOverbookedWarning/index.js
+++ b/src/lib/pages/backoffice/Loan/LoanDetails/LoanOverbookedWarning/index.js
@@ -1,0 +1,1 @@
+export { default as LoanOverbookedWarning } from './LoanOverbookedWarning';


src/lib/modules/Loan/actions.js
```diff
@@ -15,6 +15,10 @@ export const DETAILS_IS_LOADING = 'fetchLoanDetails/IS_LOADING';
 export const DETAILS_SUCCESS = 'fetchLoanDetails/SUCCESS';
 export const DETAILS_HAS_ERROR = 'fetchLoanDetails/HAS_ERROR';
 
+export const CIRCULATION_IS_LOADING = 'fetchCirculation/IS_LOADING';
```
ntarocco, line 18: You are not using Redux in the end, so you can remove this.

src/lib/pages/backoffice/Loan/LoanDetails/LoanOverbooked/LoanOverbooked.js
```diff
@@ -0,0 +1,56 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { withCancel } from '@api/utils';
+import { Segment } from 'semantic-ui-react';
+import { documentApi } from '@api/documents';
+
+export default class LoanOverbooked extends Component {
```
ntarocco, line 7: I would propose to be more specific:
```suggestion
export default class LoanOverbookedWarning extends Component {
```

src/lib/pages/backoffice/Loan/LoanDetails/LoanOverbooked/LoanOverbooked.js
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  componentWillUnmount() {
+    this.cancellableCirculation && this.cancellableCirculation.cancel();
+  }
+
+  fetchCirculation = async () => {
+    const { documentPid } = this.props;
+    try {
+      console.log(documentPid);
```
ntarocco, line 28:
```suggestion

```

src/lib/pages/backoffice/Loan/LoanDetails/LoanOverbooked/LoanOverbooked.js
```diff
@@ -0,0 +1,56 @@
...
+      this.cancellableCirculation = withCancel(
+        documentApi.get(documentPid)
+      );
+      const response = await this.cancellableCirculation.promise;
+      const circulation = response.data.metadata.circulation;
+      this.setState({ overbooked: circulation.overbooked});
+    } catch (error) {
+      console.log(error)
```
ntarocco, line 36: Missing the part when unmounting the component? Double check the other existing examples.

src/semantic-ui/site/elements/segment.overrides
```diff
@@ -66,3 +66,8 @@ Segment Overrides - REACT-INVENIO-APP-ILS
   background: @loaderDimmerColor;
 }
 
+/* backoffice */
```
ntarocco, line 69: Instead of creating new classes, why not using the [Message](https://react.semantic-ui.com/collections/message/#variations-warning) component of Semantic-UI?

src/lib/pages/backoffice/Loan/LoanDetails/LoanOverbooked/LoanOverbooked.js
```diff
@@ -0,0 +1,56 @@
...
+    }
+  };
+
+
+  render() {
+    const { overbooked } = this.state;
+    return (
+      <>
+        {overbooked && <Segment attached className="bo-metadata-segment warning-box">
+          There are people waiting to borrow the book.
```
ntarocco, line 46:
```suggestion
          This literature is <a ...>overbooked</a>.
```
The `<a href` tag should open **in a new tab** the loans search with `loans?q=document_pid:<documentPid>`

ntarocco (commented):
Added a few suggestions! :)

src/lib/pages/backoffice/Loan/LoanDetails/LoanOverbooked/LoanOverbooked.js
Pineirin: I'm not so sure if the idea would be to just hide the warning in case the component is unmounted or the error should be displayed.
Also, then I'm no so sure why in src/lib/modules/Document/DocumentStats.js there is a comment instead of a proper handling of the error.

src/lib/pages/backoffice/Loan/LoanDetails/LoanOverbookedWarning/LoanOverbookedWarning.js
```diff
@@ -0,0 +1,57 @@
...
+    try {
+      this.cancellableCirculation = withCancel(
+        documentApi.get(documentPid)
+      );
+      const response = await this.cancellableCirculation.promise;
+      const circulation = response.data.metadata.circulation;
+      this.setState({ overbooked: circulation.overbooked});
+    } catch (error) {
+      if (error !== 'UNMOUNTED') {
+        this.setState({ overbooked: false});
```
ntarocco, line 34: During yesterday stand up, @kprzerwa was suggesting to display an error message on error instead.

Example:
```javascript
OVERBOOKED_MSG = "This literature is overbooked."
OVERBOOKED_ERROR = "Cannot verify if this literature is overbooked. Please refresh the page."

export default class LoanOverbookedWarning extends Component {
...
      this.setState({ msg: circulation.overbooked ? OVERBOOKED_MSG : ''});
    } catch (error) {
      if (error !== 'UNMOUNTED') {
          this.setState({ msg: OVERBOOKED_ERROR});
```

src/lib/pages/backoffice/Loan/LoanDetails/LoanOverbookedWarning/LoanOverbookedWarning.js
```diff
@@ -0,0 +1,57 @@
...
+
+
+  render() {
+    const { overbooked } = this.state;
+    const { documentPid } = this.props;
+    return (
+      <>
+      {overbooked &&
+        <div class="ui warning message">
+          This literature is <a href={"/backoffice/loans?q=" + documentPid}>overbooked</a>.
```
ntarocco, line 47: This is not correct. Check other examples in the code on how the links are generated. Suggestion: check the `<Link to=` component.

src/lib/pages/backoffice/Loan/LoanDetails/LoanOverbooked/LoanOverbooked.js
ntarocco: > I'm not so sure if the idea would be to just hide the warning in case the component is unmounted or the error should be displayed.
When the component is `unmounted`, you cannot use `this.`, e.g. `this.setState`. We have discussed about this, you can read more for example [here](https://www.freecodecamp.org/news/how-to-work-with-react-the-right-way-to-avoid-some-common-pitfalls-fc9eb5e34d9e/).

> Also, then I'm no so sure why in src/lib/modules/Document/DocumentStats.js there is a comment instead of a proper handling of the error.
This is because in the case of document stats, we don't want to do anything at all. Imagine that you have an issue with the stats, you probably don't want to display an error message, a popup warning or banner to the user, on each book details page.

src/lib/pages/backoffice/Loan/LoanDetails/LoanDetails.js
```diff
@@ -54,6 +55,7 @@ export default class LoanDetails extends Component {
                     <Grid.Column width={13}>
                       <Container className="spaced">
                         <Container className="spaced">
+                          {data.metadata?.document_pid && <LoanOverbookedWarning documentPid={data.metadata?.document_pid} />}
```
kprzerwa, line 58:
```suggestion
                          {data.metadata?.document_pid && <LoanOverbookedWarning documentPid={data.metadata.document_pid} />}
```
you checked already if metadata exists in the first part of the statement

src/lib/pages/backoffice/Loan/LoanDetails/LoanOverbookedWarning/LoanOverbookedWarning.js
```diff
@@ -0,0 +1,57 @@
...
+  };
+
+
+  render() {
+    const { overbooked } = this.state;
+    const { documentPid } = this.props;
+    return (
+      <>
+      {overbooked &&
+        <div class="ui warning message">
```
kprzerwa, line 46: why didn't you use the Semantic UI Message component instead of the custom div ? we also have our own wrapper component to unify the look in the application - please check the available components in the code or semantic UI documentation

src/lib/pages/backoffice/Loan/LoanDetails/LoanOverbookedWarning/LoanOverbookedWarning.js
```diff
@@ -0,0 +1,73 @@
...
+    const { documentPid } = this.props;
+    return (
+      <>
+      {overbooked != OVERBOOKED_STATES.NOT_OVERBOOKED &&
+        <Message warning>
+          {overbooked != OVERBOOKED_STATES.ERROR
+            ? <>This literature is <Link to={BackOfficeRoutes.documentDetailsFor(documentPid)}><code>overbooked</code></Link>.</>
+            : <>Cannot verify if this literature is overbooked. Please refresh the page.</>
+          }
+        </Message>
```
Pineirin, line 63: I don't like this if inside an if, but I don't see any other way of implementing the <Link>.
I tried using string templates (and passing the Link object to the template) but never managed to display it properly.

src/lib/pages/backoffice/Loan/LoanDetails/LoanOverbookedWarning/LoanOverbookedWarning.js
```diff
@@ -0,0 +1,75 @@
...
+
+  fetchCirculation = async () => {
+    const { documentPid } = this.props;
+    try {
+      this.cancellableCirculation = withCancel(
+        documentApi.get(documentPid)
+      );
+      const response = await this.cancellableCirculation.promise;
+      const circulation = response.data.metadata.circulation;
+      console.log(circulation.overbooked)
```
ntarocco, line 42:
```suggestion

```

src/lib/pages/backoffice/Loan/LoanDetails/LoanOverbookedWarning/LoanOverbookedWarning.js
```diff
@@ -0,0 +1,75 @@
...
+      }
+    }
+  };
+
+
+  render() {
+    const { overbooked } = this.state;
+    const { documentPid } = this.props;
+    const link = <Link to={BackOfficeRoutes.documentDetailsFor(documentPid)}><code>overbooked</code></Link>
+    const myString = `This literature is ${link}`
```
ntarocco, line 56: What is this `myString`?

src/lib/pages/backoffice/Loan/LoanDetails/LoanOverbookedWarning/LoanOverbookedWarning.js
ntarocco: I would have implemented differently to avoid the extra ifs.

Define a couple of constants on the top:
```
OVERBOOKED_MSG = (documentPid) => <>This literature is <Link to={BackOfficeRoutes.documentDetailsFor(documentPid)}>overbooked</Link></>
ERROR_MSG = "Cannot verify if this literature is overbooked. Please refresh the page."
```

Default state in the constructor:
```
state = {
    hasError: false,
    msg: ''
}
```

When fetching the document:
```
fetchCirculation = async () => {
... if overbooked ...
this.setState({hasError: false, msg: OVERBOOKED_MSG(documentPid)});
... if error ...
this.setState({hasError: true, msg: ERROR_MSG});
```

In the render method:
```
const { hasError, msg } = this.state;
const color = hasError ? 'red' : 'yellow';
return msg && (
  <Message color={color}>
      {msg}
  </Message>
)
```

How does it look like this?

ntarocco (approved):
Very good job! I think it is much more readable now! 👏🏼

PR #463, opened 2021-07-15
Acquisition/borrowing: form fields display changes

Pineirin:
Changes to acquisitions and borrowing forms:
- Sort status values in the purchase orders form, choosing the first value as the default one.
- Select Physical copy as the default value of the type in the borrowing requests form.
- Refactor code for displaying the fields in the dropdown selectors, adding an order field and the possibility of using a default value.
closes https://github.com/CERNDocumentServer/cds-ils/issues/327
Screenshot from 2021-07-15 17-06-41
Screenshot from 2021-07-15 17-06-56

### DIFF
--- a/src/lib/config/common.js
+++ b/src/lib/config/common.js
@@ -8,52 +8,151 @@ export const DEFAULT_LANGUAGE = 'ENG';
 // text key is used for semantic ui dropdown options display
 
 export const ACQ_ORDER_STATUSES = [
-  { value: 'CANCELLED', text: 'Cancelled', label: 'Cancelled' },
-  { value: 'RECEIVED', text: 'Received', label: 'Received' },
-  { value: 'ORDERED', text: 'Ordered', label: 'Ordered' },
-  { value: 'PENDING', text: 'Pending', label: 'Pending' },
+  {
+    value: 'PENDING',
+    text: 'Pending',
+    label: 'Pending',
+    order: 1,
+    default: true,
+  },
+  {
+    value: 'ORDERED',
+    text: 'Ordered',
+    label: 'Ordered',
+    order: 2,
+    default: false,
+  },
+  {
+    value: 'RECEIVED',
+    text: 'Received',
+    label: 'Received',
+    order: 3,
+    default: false,
+  },
+  {
+    value: 'CANCELLED',
+    text: 'Cancelled',
+    label: 'Cancelled',
+    order: 4,
+    default: false,
+  },
 ];
 
 export const ILL_BORROWING_REQUESTS_STATUSES = [
-  { value: 'CANCELLED', text: 'Cancelled', label: 'Cancelled' },
-  { value: 'PENDING', text: 'Pending', label: 'Pending' },
-  { value: 'REQUESTED', text: 'Requested', label: 'Requested' },
-  { value: 'ON_LOAN', text: 'On loan', label: 'On loan' },
-  { value: 'RETURNED', text: 'Returned', label: 'Returned' },
+  {
+    value: 'CANCELLED',
+    text: 'Cancelled',
+    label: 'Cancelled',
+    order: 1,
+    default: false,
+  },
+  {
+    value: 'PENDING',
+    text: 'Pending',
+    label: 'Pending',
+    order: 2,
+    default: false,
+  },
+  {
+    value: 'REQUESTED',
+    text: 'Requested',
+    label: 'Requested',
+    order: 3,
+    default: false,
+  },
+  {
+    value: 'ON_LOAN',
+    text: 'On loan',
+    label: 'On loan',
+    order: 4,
+    default: false,
+  },
+  {
+    value: 'RETURNED',
+    text: 'Returned',
+    label: 'Returned',
+    order: 5,
+    default: false,
+  },
 ];
 
 export const ITEM_MEDIUMS = [
-  { value: 'NOT_SPECIFIED', text: 'Not specified', label: 'Not specified' },
-  { value: 'PAPER', text: 'Paper', label: 'Paper' },
-  { value: 'CDROM', text: 'CD-ROM', label: 'CD-ROM' },
-  { value: 'DVD', text: 'DVD', label: 'DVD' },
-  { value: 'VHS', text: 'VHS', label: 'VHS' },
-  { value: 'PAPERBACK', text: 'Paperback', label: 'Paperback' },
-  { value: 'HARDCOVER', text: 'Hardcover', label: 'Hardcover' },
+  {
+    value: 'NOT_SPECIFIED',
+    text: 'Not specified',
+    label: 'Not specified',
+    order: 1,
+  },
+  { value: 'PAPER', text: 'Paper', label: 'Paper', order: 2 },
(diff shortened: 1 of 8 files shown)


src/lib/config/common.js
```diff
@@ -8,10 +8,10 @@ export const DEFAULT_LANGUAGE = 'ENG';
 // text key is used for semantic ui dropdown options display
 
 export const ACQ_ORDER_STATUSES = [
-  { value: 'CANCELLED', text: 'Cancelled', label: 'Cancelled' },
-  { value: 'RECEIVED', text: 'Received', label: 'Received' },
-  { value: 'ORDERED', text: 'Ordered', label: 'Ordered' },
   { value: 'PENDING', text: 'Pending', label: 'Pending' },
+  { value: 'ORDERED', text: 'Ordered', label: 'Ordered' },
+  { value: 'RECEIVED', text: 'Received', label: 'Received' },
+  { value: 'CANCELLED', text: 'Cancelled', label: 'Cancelled' },
```
Pineirin, line 14: Looking at the code, I don't think there would be any problem using order as an attribute, but I don't see any benefit either.

src/lib/pages/backoffice/ILL/BorrowingRequest/BorrowingRequestEditor/schema.js
```diff
@@ -139,6 +139,7 @@ export const schema = () => {
       type: {
         title: 'Type',
         type: 'string',
+        default: "PHYSICAL_COPY"
```
ntarocco, line 142: If this is a vocabulary, we cannot hardcode this value: it can change per instance.

src/lib/config/common.js
kprzerwa: IMO for the code maintenance reasons it should either have the order attribute (more explicitly telling us as developers that the order matters for the users).
Minimum would be adding a comment which tells us to preserve the ordering.

src/lib/config/common.js
```diff
@@ -8,10 +8,10 @@ export const DEFAULT_LANGUAGE = 'ENG';
 // text key is used for semantic ui dropdown options display
 
 export const ACQ_ORDER_STATUSES = [
-  { value: 'CANCELLED', text: 'Cancelled', label: 'Cancelled' },
-  { value: 'RECEIVED', text: 'Received', label: 'Received' },
-  { value: 'ORDERED', text: 'Ordered', label: 'Ordered' },
-  { value: 'PENDING', text: 'Pending', label: 'Pending' },
+  { value: 'PENDING', text: 'Pending', label: 'Pending', order: 1 },
```
ntarocco, line 11: This is nice, but then you need to add the code that when rendering these statuses, you will have to sort by `order`.
Please add this below in the other statuses too.

src/lib/pages/backoffice/Acquisition/Order/OrderEditor/schema.js
```diff
@@ -196,6 +196,7 @@ export const schema = () => {
         enum: Object.values(invenioConfig.ACQ_ORDERS.statuses).map(
           (s) => s.value
         ),
+        default: Object.values(invenioConfig.ACQ_ORDERS.statuses)[0].value
```
ntarocco, line 199: Why not adding a `default` field in the statuses and use that instead of the first value?

src/lib/pages/backoffice/ILL/BorrowingRequest/BorrowingRequestEditor/schema.js
```diff
@@ -139,6 +139,7 @@ export const schema = () => {
       type: {
         title: 'Type',
         type: 'string',
+        default: invenioConfig.ILL_BORROWING_REQUESTS.defaultType
```
ntarocco, line 142: Do not forget to create a PR in CDS-ILS to override this.

kprzerwa, line 142: sorry, not quite sure how does this work. We set the defaultType to `''` (empty string)?

src/lib/pages/backoffice/Acquisition/Order/OrderEditor/schema.js
```diff
@@ -193,9 +193,10 @@ export const schema = () => {
       status: {
         title: 'Status',
         type: 'string',
-        enum: Object.values(invenioConfig.ACQ_ORDERS.statuses).map(
+        enum: Object.values(invenioConfig.ACQ_ORDERS.statuses).sort((a, b) => a.order - b.order).map(
           (s) => s.value
         ),
+        default: Object.values(invenioConfig.ACQ_ORDERS.statuses).find((s) => s.default).value
```
ntarocco, line 199: What about using [lodash _find](https://lodash.com/docs/4.17.15#find) so that if there is no `default` key it will still work?
```suggestion
        default: _get(_find(invenioConfig.ACQ_ORDERS.statuses, { 'default': true }), 'value', null)
```
to be tested! (`_find` will return `undefined` if not found). You also need to if default should have `null` or `''` when there is no default in react-jsonschema-forms.

Like this you can have only one element with `default: true`, and omit it from the others.

src/lib/config/common.js
```diff
@@ -8,18 +8,42 @@ export const DEFAULT_LANGUAGE = 'ENG';
...
+  },
 ];
 
 export const ILL_BORROWING_REQUESTS_STATUSES = [
-  { value: 'CANCELLED', text: 'Cancelled', label: 'Cancelled' },
-  { value: 'PENDING', text: 'Pending', label: 'Pending' },
-  { value: 'REQUESTED', text: 'Requested', label: 'Requested' },
-  { value: 'ON_LOAN', text: 'On loan', label: 'On loan' },
-  { value: 'RETURNED', text: 'Returned', label: 'Returned' },
+  { value: 'CANCELLED', text: 'Cancelled', label: 'Cancelled', order: 1 },
```
ntarocco, line 42: Maybe it makes sense to add `order` and `default` to also **all** lists below and then change the code where they are used to take `order` and `default` into account?
WDYT? ping also @kprzerwa

src/lib/pages/backoffice/ILL/BorrowingRequest/BorrowingRequestEditor/schema.js
ntarocco: It is maybe overridden in the overlay. To be check what is the value to use when **no default**.

Pineirin: As far as I understand, with the current implementation, it will try to use the value used as 'defaultType' in the appropriate configuration file. If it doesn't find the using the value stored in the appropriate configuration file, it will display an empty space in the drop down, almost as if there was no default value code.

src/lib/config/common.js
Pineirin: The code and configuration for using the order key was implemented in all lists of the common.js. Also, the code and configuration for using the default key was implemented in those lists that were not used in a form dedicated to update data (in other words, it was implemented for the first two lists).

ntarocco (approved):

PR #470, opened 2021-08-13
check in: highlight fixes

kprzerwa:
* check in: location highlight fixes

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,24 +5,24 @@
   "requires": true,
   "dependencies": {
     "@artsy/fresnel": {
-      "version": "1.5.0",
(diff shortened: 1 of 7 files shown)


ntarocco (approved):

ntarocco:
We already have the eslint.rc: https://github.com/inveniosoftware/react-invenio-app-ils/blob/master/.eslintrc

I think we should first why that is not working? I wouldn't add two...

ntarocco (commented):
We already have the eslint.rc: https://github.com/inveniosoftware/react-invenio-app-ils/blob/master/.eslintrc

I think we should first why that is not working? I wouldn't add two...

PR #476, opened 2021-09-16
Document&AcquisitionOrder: now pre-fills additional fields

NRodriguezcuellar:
more fields are now being prefilled in the document create and acquisition order form

closes https://github.com/CERNDocumentServer/cds-ils/issues/536

### DIFF
--- a/src/lib/pages/backoffice/Document/DocumentEditor/DocumentEditor.js
+++ b/src/lib/pages/backoffice/Document/DocumentEditor/DocumentEditor.js
@@ -17,10 +17,9 @@ export class DocumentEditor extends Component {
   constructor(props) {
     super(props);
 
-    const isEditing = !!props.match.params.documentPid;
     this.state = {
       data: {},
-      isLoading: isEditing,
+      isLoading: this.userIsEditing,
       error: {},
     };
   }
@@ -35,6 +34,10 @@ export class DocumentEditor extends Component {
   componentWillUnmount() {
     this.cancellableFetch && this.cancellableFetch.cancel();
     this.cancellableAttachDocument && this.cancellableAttachDocument.cancel();
+  }
+
+  get userIsEditing() {
+    return !!this.props.match.params.documentPid;
   }
 
   fetch = async (documentPid) => {
@@ -59,8 +62,7 @@ export class DocumentEditor extends Component {
     // delete the custom authors field so that it is not sent to backend
     delete formData._tooManyAuthorsCustomField;
 
-    const isEditing = !!documentPid;
-    return isEditing
+    return this.userIsEditing
       ? await documentApi.update(documentPid, formData)
       : await documentApi.create(formData);
   };
@@ -112,8 +114,7 @@ export class DocumentEditor extends Component {
     } = this.props;
     const { isLoading, error, data } = this.state;
 
-    const isEditing = !!documentPid;
-    if (isEditing) {
+    if (this.userIsEditing) {
       const formTitle = `Document - Edit #${documentPid}`;
 
       // when there are too many authors, it will not be possible to manually edit them
--- a/src/lib/pages/backoffice/DocumentRequest/DocumentRequestDetails/DocumentRequestSteps/DocumentRequestStepPanels/ChooseDocumentStepPanel/ChooseDocumentStepPanel.js
+++ b/src/lib/pages/backoffice/DocumentRequest/DocumentRequestDetails/DocumentRequestSteps/DocumentRequestStepPanels/ChooseDocumentStepPanel/ChooseDocumentStepPanel.js
@@ -3,7 +3,7 @@ import { goTo } from '@history';
 import { ESSelector } from '@modules/ESSelector';
 import { serializeDocument } from '@modules/ESSelector/serializer';
 import { BackOfficeRoutes } from '@routes/urls';
-import _get from 'lodash/get';
+import mapFields from '@forms/mapFields';
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 import { Button, Divider, Grid, Header, Segment } from 'semantic-ui-react';
@@ -15,41 +15,26 @@ export default class ChooseDocumentStepPanel extends Component {
   };
 
   goToCreateDocumentAndPrefill = (docReq) => {
-    const data = {};
-    const authors = _get(docReq, 'authors');
-    if (authors) {
-      data['authors'] = [{ full_name: _get(docReq, 'authors') }];
-    }
-    const journalTitle = _get(docReq, 'journal_title', '');
-    const issue = _get(docReq, 'issue', '');
-    const volume = _get(docReq, 'volume', '');
-    if (journalTitle || issue || volume) {
-      data['publication_info'] = [
-        {
-          journal_title: journalTitle,
-          journal_issue: issue,
-          journal_volume: volume,
-        },
-      ];
-    }
-    const isbn = _get(docReq, 'isbn');
-    if (isbn) {
-      data['identifiers'] = [{ scheme: 'ISBN', value: isbn }];
-    }
+    const mappings = [
+      ['authors', 'authors[0].full_name'],
+      ['journal_title', 'publication_info[0].journal_title'],
+      ['issue', 'publication_info[0].journal_issue'],
+      ['volume', 'publication_info[0].journal_volume'],
+      ['isbn', 'identifiers[0]', (value) => ({ scheme: 'ISBN', value: value })],
+      ['edition', 'edition'],
+      ['publication_year', 'publication_year', (value) => value.toString()],
+      ['publisher', 'imprint.publisher'],
+    ];
 
-    const edition = _get(docReq, 'edition');
-    if (edition) {
-      data['edition'] = edition;
-    }
-    const publicationYear = _get(docReq, 'publication_year');
-    if (publicationYear) {
-      data['publication_year'] = `${publicationYear}`; // it has to be a string
-    }
+    const prefilledData = mapFields({
+      origin: docReq,
+      mappings,
+    });
 
     const createDocumentFormData = {
       formData: {
         title: docReq.title,
-        ...data,
+        ...prefilledData,
       },
       extraData: {
         attachCreatedDocumentToDocumentRequest: true,
--- a/src/lib/pages/backoffice/DocumentRequest/DocumentRequestDetails/DocumentRequestSteps/DocumentRequestStepPanels/ChooseProviderStepPanel/ChooseProviderStepPanel.js
+++ b/src/lib/pages/backoffice/DocumentRequest/DocumentRequestDetails/DocumentRequestSteps/DocumentRequestStepPanels/ChooseProviderStepPanel/ChooseProviderStepPanel.js
@@ -12,10 +12,10 @@ import {
   serializeBorrowingRequest,
 } from '@modules/ESSelector/serializer';
 import { AcquisitionRoutes, ILLRoutes } from '@routes/urls';
-import _get from 'lodash/get';
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 import { Button, Divider, Grid, Label, Segment } from 'semantic-ui-react';
+import mapFields from '@forms/mapFields';
 
 class AcqProvider extends Component {
   onSelectResult = (provData) => {
@@ -25,18 +25,20 @@ class AcqProvider extends Component {
   };
 
   goToCreateOrderAndPrefill = (docReq) => {
+    const mappings = [
+      ['document_pid', 'document_pid'],
+      ['patron_pid', 'patron_pid'],
+      ['paymeny_info', 'budget_code'],
+      ['payment_method', 'payment_mode'],
+    ];
+
     const createOrderFormData = {
       formData: {
-        order_lines: [
-          {
-            document_pid: _get(docReq, 'document_pid', ''),
-            patron_pid: docReq.patron_pid,
-          },
-        ],
-      },
-      extraData: {
-        attachCreatedOrderToDocumentRequest: true,
-        documentRequestPid: docReq.pid,
+        order_lines: [mapFields({ mappings, origin: docReq })],
+        extraData: {
+          attachCreatedOrderToDocumentRequest: true,
+          documentRequestPid: docReq.pid,
+        },
       },
     };
 
@@ -90,11 +92,16 @@ class IllProvider extends Component {
   };
 
   goToCreateBrwReqAndPrefill = (docReq) => {
+    const mappings = [
+      ['document_pid', 'document_pid'],
+      ['patron_pid', 'patron_pid'],
+    ];
+
     const createBrwReqFormData = {
-      formData: {
-        document_pid: _get(docReq, 'document_pid', ''),
-        patron_pid: docReq.patron_pid,
-      },
+      formData: mapFields({
+        mappings,
+        origin: docReq,
+      }),
       extraData: {
         attachCreatedBrwReqToDocumentRequest: true,
         documentRequestPid: docReq.pid,
--- a/src/lib/pages/backoffice/DocumentRequest/DocumentRequestEditor/DocumentRequestEditor.js
+++ b/src/lib/pages/backoffice/DocumentRequest/DocumentRequestEditor/DocumentRequestEditor.js
@@ -14,10 +14,9 @@ export class DocumentRequestEditor extends Component {
   constructor(props) {
     super(props);
 
-    const isEditing = !!props.match.params.documentRequestPid;
     this.state = {
       data: {},
-      isLoading: isEditing,
+      isLoading: this.userIsEditing,
       error: {},
     };
   }
@@ -31,6 +30,10 @@ export class DocumentRequestEditor extends Component {
 
   componentWillUnmount() {
     this.cancellableFetchDocReq && this.cancellableFetchDocReq.cancel();
+  }
+
+  get userIsEditing() {
+    return !!this.props.match.params.documentRequestPid;
   }
 
   fetchDocReq = async (documentRequestPid) => {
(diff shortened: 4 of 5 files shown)


src/lib/pages/backoffice/Document/DocumentEditor/DocumentEditor.js
```diff
@@ -17,10 +17,9 @@ export class DocumentEditor extends Component {
   constructor(props) {
     super(props);
 
-    const isEditing = !!props.match.params.documentPid;
     this.state = {
       data: {},
-      isLoading: isEditing,
+      isLoading: this.userOnEditPage(),
```
zzacharo, line 22: Was the rename needed here?

src/lib/pages/backoffice/DocumentRequest/DocumentRequestDetails/DocumentRequestSteps/DocumentRequestStepPanels/ChooseDocumentStepPanel/ChooseDocumentStepPanel.js
```diff
@@ -15,36 +15,33 @@ export default class ChooseDocumentStepPanel extends Component {
...
-    const edition = _get(docReq, 'edition');
-    if (edition) {
-      data['edition'] = edition;
-    }
-    const publicationYear = _get(docReq, 'publication_year');
-    if (publicationYear) {
-      data['publication_year'] = `${publicationYear}`; // it has to be a string
-    }
+    const data = {
+      authors: authors && [{ full_name: authors }],
```
zzacharo, line 30: this is not correct according to the behaviour that was there before. The `authors` key now will be always there even if it's `undefined` compared to before that was added only if defined.

src/lib/pages/backoffice/DocumentRequest/DocumentRequestDetails/DocumentRequestSteps/DocumentRequestStepPanels/ChooseDocumentStepPanel/ChooseDocumentStepPanel.js
```diff
@@ -15,36 +15,33 @@ export default class ChooseDocumentStepPanel extends Component {
...
+      publication_info: _some([journalTitle, issue, volume])
```
zzacharo, line 31: I think the same goes here.

src/lib/pages/backoffice/DocumentRequest/DocumentRequestDetails/DocumentRequestSteps/DocumentRequestStepPanels/ChooseDocumentStepPanel/ChooseDocumentStepPanel.js
```diff
@@ -15,36 +15,33 @@ export default class ChooseDocumentStepPanel extends Component {
...
+            {
+              journal_title: journalTitle ?? '',
+              journal_issue: issue ?? '',
+              journal_volume: volume ?? '',
+            },
+          ]
+        : [],
+      identifiers: isbn ? [{ scheme: 'ISBN', value: isbn }] : [],
+      edition: edition,
+      publication_year: publication_year?.toString(),
```
zzacharo, line 42: See my previous comment for this one too.

src/lib/pages/backoffice/DocumentRequest/DocumentRequestDetails/DocumentRequestSteps/DocumentRequestStepPanels/ChooseDocumentStepPanel/ChooseDocumentStepPanel.js
```diff
@@ -15,36 +15,33 @@ export default class ChooseDocumentStepPanel extends Component {
...
+      imprint: { publisher: publisher ?? '' },
```
zzacharo, line 43: I would say let's match the code style and behavior that was before in place just to have some consistency on the code i.e in the way we check for values existence.

src/lib/pages/backoffice/DocumentRequest/DocumentRequestDetails/DocumentRequestSteps/DocumentRequestStepPanels/ChooseProviderStepPanel/ChooseProviderStepPanel.js
```diff
@@ -31,6 +31,8 @@ class AcqProvider extends Component {
           {
             document_pid: _get(docReq, 'document_pid', ''),
             patron_pid: docReq.patron_pid,
+            payment_mode: docReq.payment_method,
```
zzacharo, line 34: If `payment_mode` is not required, thus ensured that is always there, I would go for something like 
```javascript
{
...,
payment_mode: _get(docReq, "payment_mode", <default_value_for_field_type>),
...
}
```

src/lib/pages/backoffice/DocumentRequest/DocumentRequestDetails/DocumentRequestSteps/DocumentRequestStepPanels/ChooseProviderStepPanel/ChooseProviderStepPanel.js
```diff
@@ -31,6 +31,8 @@ class AcqProvider extends Component {
...
+            budget_code: docReq.payment_info,
```
zzacharo, line 35: Same as above.

src/lib/pages/backoffice/DocumentRequest/DocumentRequestEditor/DocumentRequestEditor.js
```diff
@@ -105,6 +105,7 @@ export class DocumentRequestEditor extends Component {
 
 DocumentRequestEditor.propTypes = {
   match: PropTypes.shape({
+    path: PropTypes.string,
```
zzacharo, line 108: Is this used somewhere?

src/lib/pages/backoffice/DocumentRequest/DocumentRequestDetails/DocumentRequestSteps/DocumentRequestStepPanels/ChooseDocumentStepPanel/ChooseDocumentStepPanel.js
NRodriguezcuellar: ah ok! I thought it would not matter much because if there is no value given it would be undefined anyhow

src/lib/pages/backoffice/Document/DocumentEditor/DocumentEditor.js
NRodriguezcuellar: I like being verbose but that maybe just my style hehe

zzacharo: I agree with verbosity! But here the name is referring to the document state and not the user action. Also, I think you don't need a function to evaluate that condition unless I am missing something!

src/lib/pages/backoffice/DocumentRequest/DocumentRequestDetails/DocumentRequestSteps/DocumentRequestStepPanels/ChooseDocumentStepPanel/ChooseDocumentStepPanel.js
zzacharo: That depends on how this value is checked on other parts of the code. Maybe @kprzerwa can confirm if including in the `data` object only the defined values are mandatory or not :)

src/lib/pages/backoffice/Document/DocumentEditor/DocumentEditor.js
NRodriguezcuellar: fair enough! then maybe  **userIsEditing** is better ? 

I made it into a function because the same logic was being repeated a lot (checking if there was a documentRequestPid in the url) and that did not seem that clean to me.

zzacharo: Ok, now I see what you meant. Then, just extract that to a `get` method i.e
```javascript
get isUserEditing = () => !!this.props.match.params.documentPid;
```
then you can use it as `if(this.isUserEditing) {...}`

NRodriguezcuellar: good one!

src/lib/pages/backoffice/DocumentRequest/DocumentRequestEditor/DocumentRequestEditor.js
NRodriguezcuellar: forgot about that one hehe

src/lib/pages/backoffice/DocumentRequest/DocumentRequestDetails/DocumentRequestSteps/DocumentRequestStepPanels/helper.js
```diff
@@ -0,0 +1,64 @@
...
+ *        all args are required although the third item can be omitted in an individual mapping array
+ *
+ *
+ * @description
+ * Function that resolves a set of arrays with args at specific indexes to output an object.
+ * Can be used to easily check if a variable exists and set its value to a specific path inside the new object or
+ * to only assign a certain value if a property exists
+ */
+
+export const resolvedMappings = (args) => {
```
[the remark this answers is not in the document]
NRodriguezcuellar: always up for that :)

src/lib/pages/backoffice/Document/DocumentEditor/DocumentEditor.js
```diff
@@ -59,8 +62,7 @@ export class DocumentEditor extends Component {
     // delete the custom authors field so that it is not sent to backend
     delete formData._tooManyAuthorsCustomField;
 
-    const isEditing = !!documentPid;
-    return isEditing
+    return this.userIsEditing
```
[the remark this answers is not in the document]
NRodriguezcuellar: i turned it into a getter so that means when this property is accessed it will return the function value without needing to put parenthesis ()

src/lib/pages/backoffice/DocumentRequest/DocumentRequestEditor/DocumentRequestEditor.js
```diff
@@ -54,8 +55,7 @@ export class DocumentRequestEditor extends Component {
       },
     } = this.props;
 
-    const isEditing = !!documentRequestPid;
-    return isEditing
+    return this.userOnEditPage()
```
kprzerwa, line 58: shouldn't this be a getter then ?

NRodriguezcuellar: you're completely right!

src/lib/pages/backoffice/Document/DocumentEditor/DocumentEditor.js
```diff
@@ -37,6 +36,10 @@ export class DocumentEditor extends Component {
     this.cancellableAttachDocument && this.cancellableAttachDocument.cancel();
   }
 
+  get userIsEditing() {
+    return !!this.props.match.params.documentPid;
```
ntarocco, line 40: Very minor: I think you have a linter issue here, shouldn't it be:
```
const {
      match: {
        params: { documentPid },
      },
} = this.props;
...
```
Did you install Prettier?

src/lib/pages/backoffice/DocumentRequest/DocumentRequestDetails/DocumentRequestSteps/DocumentRequestStepPanels/helper.js
```diff
@@ -0,0 +1,41 @@
...
+ *
+ *
+ * @description
+ * Function that checks an array of arrays containing instructions in the following format [propertyToCheck, pathToPutPropertyValue, a function to customize the value of the property]
+ *@description
+ * it will check if the property (can be nested using the lodash dot syntax) exists in the origin object and if so it will set the property value on the chosen path in the destinationObject
+ */
+
+export const extractAndCreate = (args) => {
+  const { destination, mappings, origin } = args;
```
ntarocco, line 23: Why this and not simply:
```
export const extractAndCreate = ({ destination, mappings: , origin }) => {
```
?
You can also use default values if some args are optional.

src/lib/pages/backoffice/DocumentRequest/DocumentRequestDetails/DocumentRequestSteps/DocumentRequestStepPanels/helper.js
```diff
@@ -0,0 +1,41 @@
+import _cloneDeep from 'lodash/cloneDeep';
```
ntarocco, line 1: In general, I would recommend avoiding as much as possible to create utils, helpers or similar (if you did not already, `Clean Code` is a very good reading). We do have `utils` modules, but I am always trying to challenge it.
On top of that, this helper is located in a specific component, but it is meant to be very generic and reusable.

My feeling is that this kind of functions become so abstract that are not re-used in other context but only in the context/use case that they were implemented for.

What about if we go for something a little more specific? You could assume that this function is to be used in the context of prefilling forms only, for example. Then, it could be placed inside the `forms` package/module.
You could then rename it to something more specific, e.g. `mapFields`.
Happy to chat about this :)

src/lib/pages/backoffice/DocumentRequest/DocumentRequestDetails/DocumentRequestSteps/DocumentRequestStepPanels/helper.js
```diff
@@ -0,0 +1,41 @@
...
+ * Function that checks an array of arrays containing instructions in the following format [propertyToCheck, pathToPutPropertyValue, a function to customize the value of the property]
+ *@description
+ * it will check if the property (can be nested using the lodash dot syntax) exists in the origin object and if so it will set the property value on the chosen path in the destinationObject
+ */
+
+export const extractAndCreate = (args) => {
+  const { destination, mappings, origin } = args;
+  const clonedDestination = _cloneDeep(destination);
+
+  mappings.forEach((map) => {
```
ntarocco, line 26: You are assuming that this is defined, maybe you want to a default value or a check before using it.

src/lib/pages/backoffice/DocumentRequest/DocumentRequestDetails/DocumentRequestSteps/DocumentRequestStepPanels/helper.js
```diff
@@ -0,0 +1,41 @@
...
+    const [propertyKey, destinationPath, customizedValueFunction] = map;
+    const valueOfPropertyKey = _get(origin, propertyKey);
+    const propertyExists = !!valueOfPropertyKey;
+
+    if (propertyExists) {
+      const valueToSet = customizedValueFunction
+        ? customizedValueFunction(valueOfPropertyKey, origin)
+        : valueOfPropertyKey;
+
+      _set(clonedDestination, destinationPath, valueToSet);
```
ntarocco, line 36: Why the need of `destination` and having it as param if you return it? It might be easier to clone deep the `origin`, modify it and return it at the end of the function so you don't need `destination` at all.

ntarocco (commented):
I like the refactoring idea, good job! 👏🏼  I have added some comments, we can chat about it whenever you prefer.

src/lib/pages/backoffice/DocumentRequest/DocumentRequestDetails/DocumentRequestSteps/DocumentRequestStepPanels/helper.js
NRodriguezcuellar: Yes those were my doubts too but it does  work very well for these contexts so i decided to just see what your guys' opinions are. I think having it for this specific purpose is not a bad idea at all actually to justify the abstraction layer.

src/lib/pages/backoffice/DocumentRequest/DocumentRequestDetails/DocumentRequestSteps/DocumentRequestStepPanels/helper.js
NRodriguezcuellar: good catch 😄

src/lib/pages/backoffice/DocumentRequest/DocumentRequestDetails/DocumentRequestSteps/DocumentRequestStepPanels/helper.js
NRodriguezcuellar: that was for the use cases where you might wanted to not use an empty object as destination object but if we are making this function very specific then i can simplify this to only have an map and origin param,

src/lib/pages/backoffice/DocumentRequest/DocumentRequestDetails/DocumentRequestSteps/DocumentRequestStepPanels/helper.js
NRodriguezcuellar: agree 😄

src/lib/pages/backoffice/Document/DocumentEditor/DocumentEditor.js
NRodriguezcuellar: it does complain about it but the code actually does not work if written that way which is interesting.

ntarocco: ok no prob! Thanks for checking :)

ntarocco (approved):
💯 👏🏼 🚀

PR #482, opened 2021-09-23
DocumentRequest: form state fix

NRodriguezcuellar:
FIX the document request form was not holding its state after an error from the api and thus was confusing to the user, lastly also moved the request type next to the state field as it is a required field and easy to miss at the bottom.

closes https://github.com/CERNDocumentServer/cds-ils/issues/584

### DIFF
--- a/src/lib/config/defaultConfig.js
+++ b/src/lib/config/defaultConfig.js
@@ -318,7 +318,18 @@ export const RECORDS_CONFIG = {
     },
   },
   DOCUMENT_REQUESTS: {
-    states: ['ACCEPTED', 'PENDING', 'DECLINED'],
+    states: [
+      {
+        value: 'ACCEPTED',
+      },
+      {
+        value: 'PENDING',
+        default: true,
+      },
+      {
+        value: 'DECLINED',
+      },
+    ],
     physicalItemProviders: {
       acq: { name: 'Acquisition', pid_type: 'acqoid' },
       ill: { name: 'Interlibrary loan', pid_type: 'illbid' },
--- a/src/lib/pages/backoffice/DocumentRequest/DocumentRequestEditor/DocumentRequestEditor.js
+++ b/src/lib/pages/backoffice/DocumentRequest/DocumentRequestEditor/DocumentRequestEditor.js
@@ -14,23 +14,26 @@ export class DocumentRequestEditor extends Component {
   constructor(props) {
     super(props);
 
-    const isEditing = !!props.match.params.documentRequestPid;
     this.state = {
       data: {},
-      isLoading: isEditing,
+      isLoading: this.userIsEditing,
       error: {},
     };
   }
 
   componentDidMount() {
     const { match } = this.props;
-    if (match.params.documentRequestPid) {
+    if (this.userIsEditing) {
       this.fetchDocReq(match.params.documentRequestPid);
     }
   }
 
   componentWillUnmount() {
     this.cancellableFetchDocReq && this.cancellableFetchDocReq.cancel();
+  }
+
+  get userIsEditing() {
+    return !!this.props.match.params.documentRequestPid;
   }
 
   fetchDocReq = async (documentRequestPid) => {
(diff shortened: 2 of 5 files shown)


src/lib/config/defaultConfig.js
```diff
@@ -318,7 +318,20 @@ export const RECORDS_CONFIG = {
     },
   },
   DOCUMENT_REQUESTS: {
-    states: ['ACCEPTED', 'PENDING', 'DECLINED'],
+    states: [
+      {
+        text: 'ACCEPTED',
+        default: false,
```
ntarocco, line 324: Very minor: I would keep the `default` only where it is `true` and, when used, check that if the `default` property does not exist, it will be `false`. I saw this pattern introduced recently in other places, might be a good idea to fix all the `default: false` occurrences and their usage.

src/lib/pages/backoffice/DocumentRequest/DocumentRequestEditor/DocumentRequestEditor.js
```diff
@@ -54,10 +58,19 @@ export class DocumentRequestEditor extends Component {
...
-    const isEditing = !!documentRequestPid;
-    return isEditing
-      ? await documentRequestApi.update(documentRequestPid, formData)
-      : await documentRequestApi.create(formData);
+    try {
+      return this.userIsEditing
+        ? await documentRequestApi.update(documentRequestPid, formData)
+        : await documentRequestApi.create(formData);
+    } catch (error) {
+      this.setState({
```
ntarocco, line 66: I think that this should not be handled here. This same pattern is use in all the forms (try to look for `.create(formData)` in the code).
You should try to change the wrapper cmp `RJSForm.js` in `onError` function instead so that it will be applied globally.

src/lib/pages/backoffice/DocumentRequest/DocumentRequestEditor/uiSchema.js
```diff
@@ -49,7 +49,8 @@ export const uiSchema = (title) => {
         'custom:divider': 16,
       },
       {
-        state: 8,
+        state: 4,
+        request_type: 4,
```
ntarocco, line 53: Careful with this :) I think I did not add it on purpose.
The workflow of a document request varies depending on its state. Changing its state without knowing how to change other metadata can be risky.
The state of a Document Request can be changed via actions in the backoffice UI and it should be changed via the form, because it is basically `Do it only if you know what you are doing, at your own risk`.
We allow to change the state in the form only if it does **not** affect the workflow or the state of other records.

src/lib/config/defaultConfig.js
NRodriguezcuellar: fair !

src/lib/pages/backoffice/DocumentRequest/DocumentRequestEditor/DocumentRequestEditor.js
NRodriguezcuellar: ah ok! I had the more risk averse thought of not trying to change a general component but if you say so 😎

src/lib/pages/backoffice/DocumentRequest/DocumentRequestEditor/uiSchema.js
NRodriguezcuellar: ah ok! This should only be a simple switch of places.

ntarocco: Oups! I am sorry, I didn't see it... my mistake...

src/lib/config/defaultConfig.js
```diff
@@ -318,7 +318,18 @@ export const RECORDS_CONFIG = {
     },
   },
   DOCUMENT_REQUESTS: {
-    states: ['ACCEPTED', 'PENDING', 'DECLINED'],
+    states: [
+      {
+        text: 'ACCEPTED',
```
kprzerwa, line 323: minor: 
```suggestion
        value: 'ACCEPTED',
```
(and on the ones below)
because we tend to use `text` to as a display label, while these refer to actual values, and could be misleading in the context of the whole config file. Check for example `CIRCULATION.statuses`

ntarocco (approved):
Looks nice! Thanks a lot! :) 👏🏼 🚀 💯

PR #488, opened 2021-10-07
Choose Provider: fixed order not being attached to the literature req…

NRodriguezcuellar:
…uest

FIX: by not correctly nesting a formData object the order created for a literature request was being attached to the request


kprzerwa (approved):
LGTM!
Could you change the commit message to be more compact?
fe
`document request: fix purchase order assignment`

PR #500, opened 2021-11-04
RecordsTable: Use activePage to render desired rows.

Pineirin:
Active page now works.
Allow hiding the text of the footer of the table.
Screenshot from 2021-11-08 15-39-38

### DIFF
--- a/src/lib/components/ResultsTable/ResultsTable.js
+++ b/src/lib/components/ResultsTable/ResultsTable.js
@@ -22,6 +22,7 @@ export class ResultsTable extends Component {
       headerActionComponent,
       renderEmptyResultsElement,
       showMaxRows,
+      showFooterSummary,
       ...tableProps
     } = this.props;
 
@@ -38,7 +39,14 @@ export class ResultsTable extends Component {
         <ResultsTableHeader columns={columns} />
         <ResultsTableBody
           columns={columns}
-          rows={showAllResults ? data : data.slice(0, showMaxRows)}
+          rows={
+            showAllResults
+              ? data
+              : data.slice(
+                  (currentPage - 1) * showMaxRows,
+                  currentPage * showMaxRows
+                )
+          }
         />
         <ResultsTableFooter
           allRowsNumber={totalHitsCount || data.length}
@@ -48,6 +56,7 @@ export class ResultsTable extends Component {
           currentPage={currentPage}
           paginationComponent={paginationComponent}
           columnsNumber={columns.length}
+          showFooterSummary={showFooterSummary}
         />
       </Table>
     );
@@ -121,6 +130,7 @@ ResultsTable.propTypes = {
   totalHitsCount: PropTypes.number,
   renderEmptyResultsElement: PropTypes.func,
   fixed: PropTypes.bool,
+  showFooterSummary: PropTypes.bool,
 };
 
 ResultsTable.defaultProps = {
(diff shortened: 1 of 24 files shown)


src/lib/components/ResultsTable/ResultsTableFooter.js
```diff
@@ -21,9 +22,16 @@ export default class ResultsTableFooter extends Component {
       <Table.Footer fullWidth data-test="footer">
         <Table.Row>
           <Table.HeaderCell colSpan={columnsNumber + 1} textAlign="right">
-            <span className="results-table-compact-summary">
-              Showing entries {start + 1}-{end} of {allRowsNumber}{' '}
-            </span>
+            {!invertedResults ? (
```
kprzerwa, line 25: why do we add inverted results? with what does it help?

src/lib/components/ResultsTable/ResultsTable.js
```diff
@@ -38,7 +39,14 @@ export class ResultsTable extends Component {
         <ResultsTableHeader columns={columns} />
         <ResultsTableBody
           columns={columns}
-          rows={showAllResults ? data : data.slice(0, showMaxRows)}
+          rows={
+            showAllResults
+              ? data
+              : data.slice(
+                  (currentPage - 1) * showMaxRows,
```
kprzerwa, line 46: did you test what happens if ```currentPage * showMaxRows >data.length```

Pineirin: From the javascript slice documentation.
Screenshot from 2021-11-09 13-42-38

So it should work (tested).

src/lib/components/ResultsTable/ResultsTable.js
```diff
@@ -137,4 +147,5 @@ ResultsTable.defaultProps = {
   totalHitsCount: 0,
   showMaxRows: invenioConfig.APP.DEFAULT_RESULTS_SIZE,
   renderEmptyResultsElement: null,
+  hideFooterMessage: false,
```
ntarocco, line 150: Is this prop used anywhere? I don't see it used in this PR.
For readability, I would also avoid the double negation in `if` conditions, `hide` and `false`. Why not:
```suggestion
  showFooterMessage: true,
```

Pineirin: The property is used in cds-ils (example in this PR: https://github.com/CERNDocumentServer/cds-ils/pull/656). I know we don't want to modify invenio for changes needed by cds, but the other potential solutions are quite bad; plus I think it makes a lot of sense to hide a footer that becomes redundant an even confusing once you add pagination to the table.

src/lib/components/ResultsTable/ResultsTable.js
kprzerwa: 👍

src/lib/index.js
```diff
@@ -41,3 +41,4 @@ export { withCancel, recordToPidType } from '@api/utils';
 export { Media } from '@components/Media';
 export { IEFallback } from '@components/Fallbacks/IEFallback';
 export { InfoPopup } from '@components/InfoPopup';
+export { Pagination } from '@components/Pagination';
```
ntarocco, line 44: This is now growing. For the future, ping @kprzerwa, we might want to export by component, pages, etc...

ntarocco (approved):

PR #501, opened 2021-11-09
DocumentItem: show items on loan in 'where to find' section

Pineirin:
Screenshot from 2021-11-09 11-38-46

### DIFF
--- a/src/lib/pages/frontsite/Documents/DocumentDetails/DocumentItems/DocumentItemBody.js
+++ b/src/lib/pages/frontsite/Documents/DocumentDetails/DocumentItems/DocumentItemBody.js
@@ -10,16 +10,21 @@ export default class DocumentItemBody extends Component {
   }
 
   itemStatus = (item) => ({
-    IsOnShelf: () =>
-      invenioConfig.ITEMS.canCirculateStatuses.includes(item.status) &&
-      !item.circulation,
+    canCirculate: () =>
+      invenioConfig.ITEMS.canCirculateStatuses.includes(item.status),
+    isOnShelf: () => !item.circulation,
   });
 
   statusLabel = (item) => {
-    const onShelf = this.itemStatus(item).IsOnShelf();
+    const canCirculate = this.itemStatus(item).canCirculate();
+    const onShelf = this.itemStatus(item).isOnShelf();
 
-    if (onShelf) {
-      return <span className="success">On shelf</span>;
+    if (canCirculate) {
+      return onShelf ? (
+        <span className="success">On shelf</span>
+      ) : (
+        <span>On loan</span>
+      );
     }
 
     return getDisplayVal('ITEMS.statuses', item.status);
--- a/src/lib/pages/frontsite/Documents/DocumentDetails/DocumentItems/DocumentItems.js
+++ b/src/lib/pages/frontsite/Documents/DocumentDetails/DocumentItems/DocumentItems.js
@@ -14,9 +14,8 @@ class DocumentItems extends Component {
   }
 
   itemStatus = (item) => ({
-    isOnShelf: () =>
-      invenioConfig.ITEMS.canCirculateStatuses.includes(item.status) &&
-      !item.circulation,
+    canCirculate: () =>
+      invenioConfig.ITEMS.canCirculateStatuses.includes(item.status),
 
     isForReference: () =>
       invenioConfig.ITEMS.referenceStatuses.includes(item.status),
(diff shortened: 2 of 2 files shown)


src/lib/pages/frontsite/Documents/DocumentDetails/DocumentItems/DocumentItemBody.js
```diff
@@ -10,16 +10,29 @@ export default class DocumentItemBody extends Component {
...
+      ITEM_IN_TRANSIT_FOR_PICKUP: 'In Transit For Pickup',
+      ITEM_IN_TRANSIT_TO_HOUSE: 'In Transit To House',
+    };
 
-    if (onShelf) {
-      return <span className="success">On shelf</span>;
+    const isOnCirculation = this.itemStatus(item).isOnCirculation();
+    const onShelf = this.itemStatus(item).isOnShelf();
+
+    console.log(item);
```
NRodriguezcuellar, line 29: 👀

src/lib/pages/frontsite/Documents/DocumentDetails/DocumentItems/DocumentItemBody.js
```diff
@@ -10,16 +10,28 @@ export default class DocumentItemBody extends Component {
...
   });
 
   statusLabel = (item) => {
-    const onShelf = this.itemStatus(item).IsOnShelf();
+    const circulationStates = {
+      ITEM_AT_DESK: 'At Desk',
+      ITEM_ON_LOAN: 'On Loan',
+      ITEM_IN_TRANSIT_FOR_PICKUP: 'In Transit For Pickup',
+      ITEM_IN_TRANSIT_TO_HOUSE: 'In Transit To House',
+    };
```
NRodriguezcuellar, line 24: For maintainability it is better to import objects from central files, i searched around the project and in the config/defaultConfig.js there is a records_config file with a property called "loanActiveStates" that looks pretty similar 😜

Pineirin: You are right, still, I cannot just use the defaultConfig, as I would need each element of the loanActiveStates to have a key (ej: ITEM_ON_LOAN) and a value (ej: 'On Loan'),  currently I have to manually add the value so I cannot take a lot of advantage of the config.

One possibility would be to had this object (circulationStates) to the config so it's easily reusable, but I don't even know if this is the solution that we should adopt (we could go for something simpler, like just putting 'On Loan' for everything).

src/lib/pages/frontsite/Documents/DocumentDetails/DocumentItems/DocumentItemBody.js
```diff
@@ -10,16 +10,28 @@ export default class DocumentItemBody extends Component {
   }
 
   itemStatus = (item) => ({
-    IsOnShelf: () =>
-      invenioConfig.ITEMS.canCirculateStatuses.includes(item.status) &&
-      !item.circulation,
+    isOnCirculation: () =>
```
ntarocco, line 13: Let's keep the name we normally use:
```suggestion
    canCirculate: () =>
```

src/lib/pages/frontsite/Documents/DocumentDetails/DocumentItems/DocumentItemBody.js
ntarocco: To make it simple, no need for the various circulationStates above, please just change it to:
```
if (canCirculate) {
     return onShelf ? (
        <span className="success">On shelf</span>
      ) : (
        <span>On loan</span>
      );
```
Should we give it a color when on loan ? orange maybe? ping @kprzerwa

src/lib/pages/frontsite/Documents/DocumentDetails/DocumentItems/DocumentItems.js
```diff
@@ -14,9 +14,8 @@ class DocumentItems extends Component {
   }
 
   itemStatus = (item) => ({
-    isOnShelf: () =>
-      invenioConfig.ITEMS.canCirculateStatuses.includes(item.status) &&
-      !item.circulation,
+    isOnCirculation: () =>
```
ntarocco, line 17:
```suggestion
    canCirculate: () =>
```

src/lib/pages/frontsite/Documents/DocumentDetails/DocumentItems/DocumentItemBody.js
Pineirin: Had a small chat with @kprzerwa and @zzacharo. The final conclusion is to keep it grey.

kprzerwa (approved):

ntarocco (approved):

PR #502, opened 2021-11-09
RJSFVocabulary: validate value

NRodriguezcuellar:
Now validates if the value passed to a vocabulary field is a valid option of that vocabulary

Screenshot 2021-11-15 at 15 04 07

closes https://github.com/CERNDocumentServer/cds-ils/issues/653

### DIFF
--- a/src/lib/forms/rjsf/widgets/RJSFVocabulary.js
+++ b/src/lib/forms/rjsf/widgets/RJSFVocabulary.js
@@ -30,6 +30,18 @@ export class RJSFVocabulary extends Component {
     this.cancellableFetchData && this.cancellableFetchData.cancel();
   }
 
+  validateValue = (value, options, isLoading) => {
+    const noOptionSelected = !value;
+    if (noOptionSelected) return true;
+
+    if (!options.length || isLoading) return;
+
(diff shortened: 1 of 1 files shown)


src/lib/forms/rjsf/widgets/RJSFVocabulary.js
```diff
@@ -82,13 +91,14 @@ export class RJSFVocabulary extends Component {
...
     const { options, isLoading, value, error } = this.state;
+
     return (
       <Form.Select
         fluid
         selection
         options={options}
         label={label}
-        value={value}
+        value={this.validateValue(value, options) ? value : undefined}
```
kprzerwa, line 101: did you check the documentation of RJSF? they have a prop responsible for validation function

NRodriguezcuellar: ah ok, that escaped my mind

kprzerwa: yes, true. However in this case we should display an error not simply remove the value

ntarocco (approved):

PR #503, opened 2021-11-09
HttpError: export base component

NRodriguezcuellar:
Export the base HttpError component for customizability

### DIFF
--- a/src/lib/components/HttpErrors/index.js
+++ b/src/lib/components/HttpErrors/index.js
@@ -5,3 +5,5 @@ export {
   TooManyRequests,
   InternalServerError,
 } from './HttpErrors';
+
+export { default as HttpError } from './HttpError';
--- a/src/lib/index.js
+++ b/src/lib/index.js
@@ -35,7 +35,7 @@ export { default as InvenioILSApp } from './App';
 export { default as history } from './history';
 export { configureStore, injectAsyncReducer } from './store';
 export { ILSStore } from './App';
-export { NotFound } from '@components/HttpErrors';
+export { NotFound, HttpError } from '@components/HttpErrors';
 export { ResultsTable } from '@components/ResultsTable/ResultsTable';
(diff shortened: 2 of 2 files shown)


src/lib/components/HttpErrors/index.js
```diff
@@ -5,3 +5,5 @@ export {
   TooManyRequests,
   InternalServerError,
 } from './HttpErrors';
+
```
kprzerwa, line 8: couldn't it be moved in the same export as the ones on top?

src/lib/components/HttpErrors/HttpError.js
```diff
@@ -4,7 +4,7 @@ import { Grid, Button, Icon } from 'semantic-ui-react';
 import PropTypes from 'prop-types';
 import Overridable from 'react-overridable';
 
-class HttpError extends Component {
+export class HttpError extends Component {
```
kprzerwa, line 7: the class is already exported from this file (check the last line of this file)

NRodriguezcuellar: i know, but it does not work me if i do not export it like this

kprzerwa: I understand but it is not a correct solution, double export is misleading

src/lib/components/HttpErrors/index.js
```diff
@@ -5,3 +5,5 @@ export {
...
+export { HttpError } from './HttpError';
```
kprzerwa, line 9: did you try 
```suggestion
export { default as HttpError } from './HttpError';
```

NRodriguezcuellar: lets try

NRodriguezcuellar: looks like that does the trick

ntarocco (approved):

PR #506, opened 2021-11-16
Loans: loan already requested modal removed from first loan.

Pineirin:
Loan already requested modal added to the rest of loans.
closes: https://github.com/CERNDocumentServer/cds-ils/issues/664
implements https://github.com/inveniosoftware/react-invenio-app-ils/pull/507
Screenshot from 2021-11-16 11-12-12
Now the modal appears for the second loan (and not the first). Hovering in the link of the modal we can see it Links correctly to the first loan.
Screenshot from 2021-11-16 11-12-35

### DIFF
--- a/src/lib/api/loans/loan.js
+++ b/src/lib/api/loans/loan.js
@@ -241,8 +241,13 @@ class QueryBuilder {
     return this;
   }
 
-  sortByRequestStartDate() {
+  sortByRequestStartDateDesc() {
     this.sortBy = `&sort=-request_start_date`;
+    return this;
+  }
+
+  sortByRequestStartDateAsc() {
+    this.sortBy = `&sort=request_start_date`;
     return this;
   }
 
--- a/src/lib/modules/Loan/actions.js
+++ b/src/lib/modules/Loan/actions.js
@@ -37,7 +37,7 @@ export const fetchLoanDetails = (
             .query()
             .withDocPid(response.data.metadata.document_pid)
             .withState(invenioConfig.CIRCULATION.loanRequestStates)
-            .sortByRequestStartDate()
+            .sortByRequestStartDateAsc()
             .qs()
         );
         if (pendingLoansResponse.data.total > 0) {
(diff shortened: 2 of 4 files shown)


ntarocco:
Wrong push?

Pineirin:
Explanation of the change.
The bug was caused because the function used to order by start date was giving the results inverted.
This caused undesired results like, for example, that the list were not properly ordered. https://github.com/CERNDocumentServer/cds-ils/issues/627
Also, like in the related issue, it can lead to a warning modal appearing were it shouldn't a not appearing where it should. This is due to the last loan being identified as the first, so if you try to checkout any other loan (like the first) it will tell you that there is a preceding one, and if you try to checkout this last loan it would not warn you, as the logic would detect it as the first.
The change was implemented as part of the solution in https://github.com/inveniosoftware/react-invenio-app-ils/pull/507, but I wanted to create this PR to show what was the bug and how it was solved.

src/lib/api/loans/loan.js
```diff
@@ -242,7 +242,7 @@ class QueryBuilder {
   }
 
   sortByRequestStartDate() {
-    this.sortBy = `&sort=-request_start_date`;
+    this.sortBy = `&sort=request_start_date`;
```
kprzerwa, line 245: there is a couple of problems I can see here.
1. this function can be reused somewhere else. By changing it in here you are changing it everywhere else. I would test what is being affected.
2. the name of the method is misleading, as you can see someone used it for sorting by decreasing date
3. since the sorting has an impact for displaying a modal (or not) for the proceeding loan request I would create a custom method (or even better, I would handle the sorting in the component which introduces the dependency with sorting and the modal)
for ex:

```suggestion
sortByRequestStartDateDesc() {
    this.sortBy = `&sort=-request_start_date`;
}

sortByRequestStartDateAsc() {
    this.sortBy = `&sort=request_start_date`;
}

sortToDetectPrecedingRequest() {
    this.sortBy = `&sort=request_start_date`;
}
```

Pineirin: I prefer to discuss this IRL.
For clarity I did check if it was used elsewhere (it wasn't) but I agree that naming was confusing. I would merge this PR https://github.com/inveniosoftware/react-invenio-app-ils/pull/507 before changing this, as they affect the same area.

kprzerwa (approved):

PR #508, opened 2021-11-17
LoanActions: warn the users that extend an overbooked loan.

Pineirin:
Fix the cancel modal of loan actions.
closes: https://github.com/CERNDocumentServer/cds-ils/issues/663
Screenshot from 2021-11-17 11-09-04

PR #512, opened 2021-12-01
LoanActions: warn the users that extend an overbooked loan.

Pineirin:
Fix the cancel modal of loan actions.
closes: CERNDocumentServer/cds-ils#663
re-implements: https://github.com/inveniosoftware/react-invenio-app-ils/pull/508

Modal:
Screenshot from 2021-12-01 15-02-13

After successfully pressing extend:
Screenshot from 2021-12-01 15-02-20

After unsuccessfully pressing extend:
Screenshot from 2021-12-01 15-02-35

### DIFF
--- a/src/lib/modules/Loan/actions.js
+++ b/src/lib/modules/Loan/actions.js
@@ -3,6 +3,7 @@ import { searchReady } from '@api/utils';
 import {
   sendErrorNotification,
   sendSuccessNotification,
+  addNotification,
 } from '@components/Notifications';
 import { fetchItemDetails } from './../../pages/backoffice/Item/ItemDetails/state/actions';
 import { invenioConfig } from '@config';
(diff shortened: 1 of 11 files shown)


src/lib/modules/Loan/actions.js
```diff
@@ -103,7 +104,8 @@ export const performLoanAction = (
         type: ACTION_HAS_ERROR,
         payload: error,
       });
-      dispatch(sendErrorNotification(error));
+      // TODO: Fix sendErrorNotification and use it here.
+      dispatch(addNotification('Error', error.message, 'error'));
```
kprzerwa, line 108: for the reviewers:
using `sendErrorNotification` here resulted on a button without any error notification in case of API errors on the response
IMO the usage was wrong here (`sendErrorNotification` redirects to error page instead of informing about the error, which might be confusing while performing a loan action) - so potentially the comment above could be remove, lets see if the rest of the team agrees

ntarocco: I think this has to be understood before removing it. `sendErrorNotification` is used extensively, there must be something wrong here?
If @Pineirin cannot figure it out, I will have to investigate...

kprzerwa (approved):

PR #514, opened 2021-12-06
RSJF: Fix bug on enter press

NRodriguezcuellar:
Now disables the onEnter event on RJSF inputs because it was causing some odd behaviour with empty fields spawning out of nowhere when pressing enter.

closes https://github.com/CERNDocumentServer/cds-ils/issues/673

### DIFF
--- a/src/lib/forms/rjsf/RJSForm.js
+++ b/src/lib/forms/rjsf/RJSForm.js
@@ -56,6 +56,25 @@ export class RJSForm extends Component {
     };
   }
 
+  componentDidMount() {
+    const inputs = document.getElementsByTagName('input');
+
+    this.ignoreEnterEvent(inputs);
+  }
+
+  ignoreEnterEvent = (inputElements) => {
+    /* For unknown reasons empty fields are spawned when pressing
+    the 'enter' key inside a RJSF input element so we have to manually prevent it*/
+
+    for (const input of inputElements) {
+      const isRJSFInput = input.id.includes('root');
+      if (isRJSFInput) {
+        input.onkeydown = (event) =>
+          event.key === 'Enter' ? event.preventDefault() : null;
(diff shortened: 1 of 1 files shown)


src/lib/forms/rjsf/RJSForm.js
```diff
@@ -56,6 +56,18 @@ export class RJSForm extends Component {
...
   }
 
+  componentDidMount() {
+    const inputs = document.getElementsByTagName('input');
+
+    for (let input of inputs) {
+      const isRJSFInput = input.id.includes('root');
+      if (isRJSFInput) {
+        input.onkeydown = (event) =>
+          event.key === 'Enter' ? event.preventDefault() : null;
```
zzacharo, line 66: Does returning `null` fallback to the default behavior?

NRodriguezcuellar: good one! From what i understand, it does, because the original event is not overridden/prevented (as is the case with preventDefault())

src/lib/forms/rjsf/RJSForm.js
```diff
@@ -56,6 +56,18 @@ export class RJSForm extends Component {
     };
   }
 
+  componentDidMount() {
+    const inputs = document.getElementsByTagName('input');
+
+    for (let input of inputs) {
```
ntarocco, line 62:
```suggestion
    for (const input of inputs) {
```

src/lib/forms/rjsf/RJSForm.js
NRodriguezcuellar: sure! 

Yes i did, everything works as normal :smiley:

zzacharo (approved):

kprzerwa (approved):

PR #515, opened 2021-12-10
Document details: various ui/ux modifications

NRodriguezcuellar:
- changed titles
  - Series -> Series number 
  - Publications -> Published in 
- removed the 'other' tab and placed the content in other tabs
  -  Alternative titles -> Details
  -  Alternative abstracts -> Content
- Fixed conference year not showing 
- Fixed licenses not being clickable 
- Fixed the line breaking when having multiple serials or multiparts
- Request an e-book is now a big grey button
- Statistics are now only shown if authenticated 

DESKTOP ->

image

image

image

image

image

image

image

image

MOBILE ->

Capture d’écran 2022-01-11 à 16 33 27
Capture d’écran 2022-01-11 à 16 33 10
Capture d’écran 2022-01-11 à 17 19 52

Capture d’écran 2022-01-11 à 16 28 30
Capture d’écran 2022-01-11 à 16 32 58

closes https://github.com/CERNDocumentServer/cds-ils/issues/526


src/lib/forms/rjsf/RJSForm.js
```diff
@@ -56,6 +56,25 @@ export class RJSForm extends Component {
     };
   }
 
+  componentDidMount() {
+    const inputs = document.getElementsByTagName('input');
+
+    this.ignoreEnterEvent(inputs);
+  }
+
+  ignoreEnterEvent = (inputElements) => {
```
kprzerwa, line 65: wasn't this part of one of the previous PRs?

src/lib/modules/Document/DocumentConference.js
```diff
@@ -22,22 +22,12 @@ export class DocumentConference extends Component {
...
-                    )}
-                </Table.Cell>
-              </Table.Row>
               <Table.Row>
                 <Table.Cell>Place</Table.Cell>
                 <Table.Cell>{conf.place}</Table.Cell>
               </Table.Row>
               <Table.Row>
-                <Table.Cell>Series</Table.Cell>
+                <Table.Cell>Series Number</Table.Cell>
```
kprzerwa, line 30:
```suggestion
                <Table.Cell>Series number</Table.Cell>
```

src/lib/modules/Document/DocumentContent.js
```diff
@@ -31,6 +36,16 @@ export class DocumentContent extends Component {
 
         <Divider horizontal>Subject classification</Divider>
         <DocumentSubjects metadata={metadata} />
+
+        <Divider horizontal>Alternative Abstracts</Divider>
```
kprzerwa, line 40:
```suggestion
        <Divider horizontal>Alternative abstracts</Divider>
```

src/lib/modules/Identifiers/Identifiers.js
```diff
@@ -70,6 +169,29 @@ export const IdentifierRows = ({ includeSchemes, identifiers }) => {
   });
 };
 
+const createValueObject = (vocabularyIdentifiers, value, scheme, material) => {
```
kprzerwa, line 172: this is CDS specific data, it should be implemented in the overlay, where the custom vocab is added

src/lib/forms/rjsf/RJSForm.js
NRodriguezcuellar: it seems i may have accidentally based this branch on that PR yes :sweat_smile:

src/lib/modules/Document/DocumentCopyrights.js
```diff
@@ -11,8 +11,8 @@ export class DocumentCopyrights extends Component {
       <List bulleted>
         {copyrights.map((entry) => (
           <List.Item key={entry.statement}>
-            <a href={entry.url}>{entry.statement}</a> ({entry.material}) held by{' '}
-            {entry.holder}
+            <a href={entry.url}>{entry.statement}</a>{' '}
+            {entry.material && `(${entry.material})`} held by {entry.holder}
```
zzacharo, line 15: in which cases is that `undefined`?

src/lib/modules/Document/DocumentExtras.js
```diff
@@ -57,9 +57,7 @@ export class DocumentExtras extends Component {
           )}
         </>
       );
-    } else {
-      return 'No additional information.';
-    }
+    } else return null;
```
zzacharo, line 60: I think you can rewrite that to something like `(!_isEmpty(metadata.alternative_titles) ||
      !_isEmpty(metadata.alternative_abstracts) )&& <>....</>`

src/lib/modules/Document/DocumentInfo.js
```diff
@@ -58,6 +72,10 @@ export class DocumentInfo extends Component {
               <Table.Cell width={4}>Title</Table.Cell>
               <Table.Cell>{metadata.title}</Table.Cell>
             </Table.Row>
+            <Table.Row>
+              <Table.Cell width={4}>Alternative titles</Table.Cell>
+              {this.renderAlternativeTitles(metadata.alternative_titles)}
```
zzacharo, line 77: I guess `alternative_titles` can be undefined so I would do something like `this.renderAlternativeTitles(_get('metadata.alternative_titles', []))` or `this.renderAlternativeTitles(metadata.alternative_titles || [])`  and remove the `isValidArray` check. That way we always expect an array or error otherwise.

src/lib/modules/Document/DocumentPublicationInfo.js
```diff
@@ -19,43 +20,43 @@ export class DocumentPublicationInfo extends Component {
                 {idx > 0 && <Divider horizontal />}
                 <Table definition key={publication.journal_title}>
                   <Table.Body>
-                    {!_isEmpty(publication.journal_title) && (
+                    {publication?.journal_title && (
```
zzacharo, line 23: Can the `_isEmpty` function be removed from the import or is it used somewhere else in this file?

src/lib/modules/Literature/LiteratureRelations.js
```diff
@@ -24,50 +24,53 @@ export default class LiteratureRelations extends Component {
   renderMultiparts = () => {
     const relations = _get(this.relations, 'multipart_monograph', []);
     if (!relations.length) return null;
+    const prefix = 'This is part of the monograph:';
```
zzacharo, line 27: I guess this can be exposed as a configurable property of the component. Just to keep things consistent with the rest of the module.

src/lib/modules/Literature/LiteratureRelations.js
```diff
@@ -24,50 +24,53 @@ export default class LiteratureRelations extends Component {
...
-        : rel.record_metadata.title;
-      return (
-        <React.Fragment key={rel.pid_value}>
-          <Link to={this.getLinkTo(rel)}>
-            <LiteratureTitle title={text} />
-          </Link>
-        </React.Fragment>
-      );
-    });
+    const prefix = 'This is part of the periodical: ';
```
zzacharo, line 54: Same here.

src/lib/pages/frontsite/Documents/DocumentDetails/DocumentDetails.js
```diff
@@ -104,11 +104,17 @@ const DocumentDetailsLayout = ({
             </ILSParagraphPlaceholder>
           </Container>
         </Container>
-        <Container textAlign="center">
-          {!_isEmpty(documentDetails.metadata) && (
-            <DocumentStats document={documentDetails} />
+        <AuthenticationGuard
```
ntarocco, line 107: Are you sure about this? I think that if you are anonymous, it will try to redirect?

NRodriguezcuellar: that is the default behaviour but i skip that by using the loginComponent prop
```
loginComponent={() => null}
```

src/lib/modules/Document/DocumentCopyrights.js
NRodriguezcuellar: material is not a required field of a Document's "copyright" property

src/lib/modules/Document/DocumentExtras.js
NRodriguezcuellar: Thanks for bringing this up, i actually can remove this tab as the information is placed in different tabs

src/lib/modules/Document/DocumentPublicationInfo.js
NRodriguezcuellar: it is :smile:

src/lib/modules/Literature/LiteratureRelations.js
NRodriguezcuellar: what do you mean by that?

src/lib/pages/frontsite/Documents/DocumentDetails/DocumentDetails.js
ntarocco: yes indeed but the code is doing: 
```
if (isAnonymous) {
      return LoginComponent ? (
        <LoginComponent />
      ) : (
        <Redirect
```
so it should go in the `Redirect` branch?

NRodriguezcuellar: well 
```
() => null; 
```
Evaluates to false so it would just call an empty object <LoginComponent /> which basically does 'nothing', which is intended compared to redirecting somewhere.

src/lib/modules/Literature/LiteratureRelations.js
NRodriguezcuellar: The component LiteratureRelations is on the 'top' components and i don't see any other top component module accepting other than props than just 'data'. So i think keeping it consistent would be keeping it like this?

src/lib/modules/Document/DocumentInfo.js
```diff
@@ -58,6 +70,10 @@ export class DocumentInfo extends Component {
...
```
kprzerwa, line 75: what will be displayed if there is no alternative titles?

In general I would like to avoid creating component methods and try to keep JSX in the render function when possible. Another way would be creating a functional component (which might already exist for displaying a list) ?

src/lib/modules/Literature/LiteratureRelations.js
kprzerwa: maybe it would make sense to refactor it into a unified functional component ? `renderMultiparts` and `renderSerials` are basically the same

src/lib/modules/Literature/LiteratureRelations.js
```diff
@@ -24,50 +24,53 @@ export default class LiteratureRelations extends Component {
...
 
     return (
       <List.Item>
-        <SeparatedList
```
kprzerwa, line 65: is SeparatedList component used anywhere else? Wouldn't it be broken visually also in the other places?

src/lib/pages/frontsite/Documents/DocumentDetails/DocumentCirculation/DocumentEItems.js
```diff
@@ -56,8 +56,13 @@ export class DocumentEItems extends Component {
...
         ) : (
           <>
-            No e-resources currently available. <br />
-            <Link to={this.onClickEItemRequestLink()}>Request the e-book.</Link>
+            <p>No e-resources currently available.</p>
+            <Link
+              to={this.onClickEItemRequestLink()}
+              className="default-margin-top"
+            >
+              <Button fluid>Request the e-book</Button>
```
kprzerwa, line 64: I think there is a shorter way ```Link as={Button}``` (to be checked)

kprzerwa (commented):
I would suggest editing the commit message fe.
```document: details page improvements
* short description 1
* short description 2
....
```
to be more detailed on the changes

src/lib/modules/Document/DocumentInfo.js
NRodriguezcuellar: just an empty list item, same as the rest in that section.

src/lib/modules/Literature/LiteratureRelations.js
NRodriguezcuellar: yes it used still even in the same component, i think for longer texts it is easy to break but otherwise it works fine.

src/lib/pages/frontsite/Documents/DocumentDetails/DocumentCirculation/DocumentEItems.js
NRodriguezcuellar: From the documentation it does not seem that way https://v5.reactrouter.com/web/api/Link

src/lib/modules/Literature/LiteratureRelations.js
```diff
@@ -13,6 +13,11 @@ export default class LiteratureRelations extends Component {
   constructor(props) {
     super(props);
     this.relations = props.relations;
+
+    this.prefixes = {
+      serial: 'This is part of the periodical: ',
+      multipart_monograph: 'This is part of the monograph:',
```
ntarocco, line 19: One has a leading space at the end, but not the other :)

ntarocco (commented):
From a quick view, LGTM :)

src/lib/pages/frontsite/Documents/DocumentDetails/DocumentCirculation/DocumentEItems.js
kprzerwa: then it was Button as={Link}, we have a lot of usage of these in the code

src/lib/modules/Document/DocumentConference.js
```diff
@@ -22,22 +22,12 @@ export class DocumentConference extends Component {
                 <Table.Cell>Dates</Table.Cell>
                 <Table.Cell>{conf.dates}</Table.Cell>
               </Table.Row>
-              <Table.Row>
```
kprzerwa, line 25: is it moved somewhere ? or just removed? If removed completely, then it is not entirely what was needed, since this is general case, it should be only removed from cds-ils UI

src/lib/pages/frontsite/Documents/DocumentDetails/DocumentPanel/DocumentPanel.js
```diff
@@ -66,15 +67,12 @@ class DocumentPanel extends Component {
...
-                            Published by{' '}
-                            <b>{doc.metadata.imprint?.publisher}</b>{' '}
-                            {doc.metadata.publication_year && (
-                              <b>{doc.metadata.publication_year}</b>
-                            )}
-                          </div>
-                        </>
+                        <div className="default-margin-bottom">
+                          Published by <b>{doc.metadata.imprint?.publisher}</b>{' '}
+                          {doc.metadata.publication_year && (
```
kprzerwa, line 72: minor: if I remember correctly publication_year is required record property so probably this check is redundant

src/lib/modules/Document/DocumentInfo.js
```diff
@@ -58,6 +70,12 @@ export class DocumentInfo extends Component {
               <Table.Cell width={4}>Title</Table.Cell>
               <Table.Cell>{metadata.title}</Table.Cell>
             </Table.Row>
+            <Table.Row>
+              <Table.Cell width={4}>Alternative titles</Table.Cell>
+              <Table.Cell>
+                {this.renderAlternativeTitles(metadata.alternative_titles)}
```
kprzerwa, line 76:
```suggestion
                <List bulleted>
        {titles.map((entry) => (
          <List.Item key={entry.value}>
            <List.Content>{entry.value}</List.Content>
          </List.Item>
        ))}
      </List>
```
I would keep the JSX in the render method, especially because we are not gaining much by separating 5 lines of the code to new method

src/lib/modules/Document/DocumentConference.js
```diff
@@ -2,48 +2,55 @@ import _isEmpty from 'lodash/isEmpty';
...
-                <Table.Cell>Series</Table.Cell>
-                <Table.Cell>{conf.series}</Table.Cell>
-              </Table.Row>
-            </Table.Body>
-          </Table>
-          <br />
-        </>
+        <Overridable
+          id="DocumentConference.Table"
+          key={conf.title + conf.acronym}
```
kprzerwa, line 13: what if someone adds two equal conferences ? the key will be invalid

NRodriguezcuellar: That is true!

src/lib/modules/Document/DocumentInfo.js
NRodriguezcuellar: fair!

src/lib/pages/frontsite/Documents/DocumentDetails/DocumentPanel/DocumentPanel.js
NRodriguezcuellar: none of the publications fields are required from what i can see!

kprzerwa:
I would suggest working a bit on the presentation of the alternative titles in the mobile view. There is a bit of problem with the white spaces around the list elements. It can be done in a separate PR, let me know if you prefer to create a new one

src/lib/pages/frontsite/Documents/DocumentDetails/DocumentPanel/DocumentPanel.js
kprzerwa: publication year is not part of the imprint but it is on the parent level. Did you check the json schema required fields?

NRodriguezcuellar: ah my bad, you are right

kprzerwa (approved):

PR #518, opened 2022-01-05
PatronProfile: add pagination element at the beginning of the lists

Pineirin:
Add information about the total loans / document requests of each list.
Allow showing all loans / document requests.
Allow revert to using pagination.
Fix pagination in lists that display document requests.
closes CERNDocumentServer/cds-ils#672

Current Loans:
Screenshot from 2022-01-12 11-28-19
Screenshot from 2022-01-12 11-28-33
Screenshot from 2022-01-12 11-28-44

Current Loans show all:
Screenshot from 2022-01-12 11-28-59
Screenshot from 2022-01-12 11-29-07

Pending Loans
Screenshot from 2022-01-12 11-29-20
Screenshot from 2022-01-12 11-29-31

Pending Loans show all:
Screenshot from 2022-01-12 11-29-43
Screenshot from 2022-01-12 11-29-54

Requests for new literature:
Screenshot from 2022-01-12 11-30-03
Screenshot from 2022-01-12 12-13-45

Requests for new Literature show all:
Screenshot from 2022-01-12 11-32-19

Past Loans:
Screenshot from 2022-01-12 11-32-34
Screenshot from 2022-01-12 11-32-46
Screenshot from 2022-01-12 11-32-56

Past Requests for new Literature:
Screenshot from 2022-01-12 11-33-07

### DIFF
--- a/src/lib/config/defaultConfig.js
+++ b/src/lib/config/defaultConfig.js
@@ -86,6 +86,7 @@ export const APP_CONFIG = {
   },
   HOME_SEARCH_BAR_PLACEHOLDER:
     'Search for books, ebooks, series, journals and standards.',
+  PATRON_PROFILE_MAX_RESULTS_SIZE: 100,
 };
 
 export const RECORDS_CONFIG = {
--- a/src/lib/pages/frontsite/PatronProfile/LoansList.js
+++ b/src/lib/pages/frontsite/PatronProfile/LoansList.js
@@ -1,21 +1,31 @@
-import { Pagination } from '@components/Pagination';
+import { PatronPagination } from './PatronPagination';
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
-import { Container, Item } from 'semantic-ui-react';
+import { Item } from 'semantic-ui-react';
 
 export default class LoansList extends Component {
   render() {
     const {
-      activePage,
-      isLoading,
-      onPageChange,
       loans,
-      rowsPerPage,
       renderListEntry,
       noLoansCmp,
+      patronShowLink,
+      onPageChange,
+      activePage,
+      rowsPerPage,
+      isLoading,
     } = this.props;
+
     return loans.total > 0 ? (
       <>
+        <PatronPagination
+          currentPage={activePage}
+          currentSize={rowsPerPage}
+          loading={isLoading}
+          onPageChange={onPageChange}
+          items={loans}
+        />
+        {patronShowLink}
         <Item.Group divided>
           {loans.hits.map((loan) => (
             <React.Fragment key={loan.metadata.pid}>
@@ -23,15 +33,13 @@ export default class LoansList extends Component {
             </React.Fragment>
           ))}
         </Item.Group>
-        <Container textAlign="center">
-          <Pagination
-            currentPage={activePage}
-            currentSize={rowsPerPage}
-            loading={isLoading}
-            onPageChange={onPageChange}
-            totalResults={loans.total}
-          />
-        </Container>
+        <PatronPagination
+          currentPage={activePage}
+          currentSize={rowsPerPage}
+          loading={isLoading}
+          onPageChange={onPageChange}
+          items={loans}
+        />
       </>
     ) : (
       noLoansCmp
@@ -47,4 +55,9 @@ LoansList.propTypes = {
   rowsPerPage: PropTypes.number.isRequired,
   renderListEntry: PropTypes.func.isRequired,
   noLoansCmp: PropTypes.element.isRequired,
+  patronShowLink: PropTypes.element,
+};
+
+LoansList.defaultProps = {
+  patronShowLink: null,
 };
--- a/src/lib/pages/frontsite/PatronProfile/PatronCurrentDocumentRequests/PatronCurrentDocumentRequests.js
+++ b/src/lib/pages/frontsite/PatronProfile/PatronCurrentDocumentRequests/PatronCurrentDocumentRequests.js
@@ -4,7 +4,6 @@ import { searchReady, withCancel } from '@api/utils';
 import { Error } from '@components/Error';
 import { ILSItemPlaceholder } from '@components/ILSPlaceholder/ILSPlaceholder';
 import { InfoMessage } from '@components/InfoMessage';
-import { Pagination } from '@components/Pagination';
 import { ResultsTable } from '@components/ResultsTable/ResultsTable';
 import LiteratureTitle from '@modules/Literature/LiteratureTitle';
 import { FrontSiteRoutes } from '@routes/urls';
@@ -13,15 +12,11 @@ import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 import Overridable from 'react-overridable';
 import { Link } from 'react-router-dom';
-import {
-  Button,
-  Container,
-  Grid,
-  Header,
-  Message,
-  Popup,
-} from 'semantic-ui-react';
+import { Button, Grid, Header, Message, Popup } from 'semantic-ui-react';
 import PatronCancelModal from '../PatronCancelModal';
+import { PatronShowLink } from '../PatronShowLink';
+import { PatronPagination } from '../PatronPagination';
+import { invenioConfig } from '@config';
 
 class ButtonCancelRequest extends Component {
   constructor(props) {
@@ -233,9 +228,22 @@ class PatronCurrentDocumentRequests extends Component {
     this.setState({ isSuccessMessageVisible: false, successMessage: '' });
   };
 
+  onShowAll = () => {
+    const { patronPid, fetchPatronDocumentRequests } = this.props;
+    fetchPatronDocumentRequests(patronPid, {
+      page: 1,
+      size: invenioConfig.APP.PATRON_PROFILE_MAX_RESULTS_SIZE,
+    });
+  };
+
+  onShowLess = () => {
+    this.fetchPatronDocumentRequests();
+  };
+
   render() {
     const { documentRequests, isLoading, error, rowsPerPage } = this.props;
     const { activePage, isSuccessMessageVisible, successMessage } = this.state;
+    const headerTitle = `Your requests for new literature (${documentRequests.total})`;
     const columns = this.getColumns();
     return (
       <Overridable
@@ -248,7 +256,7 @@ class PatronCurrentDocumentRequests extends Component {
         <>
           <Header
             as="h2"
-            content="Your requests for new literature"
+            content={headerTitle}
             className="highlight"
             textAlign="center"
           />
@@ -269,6 +277,19 @@ class PatronCurrentDocumentRequests extends Component {
           )}
           <ILSItemPlaceholder fluid isLoading={isLoading}>
             <Error error={error}>
+              <PatronPagination
+                currentPage={activePage}
+                currentSize={rowsPerPage}
+                loading={isLoading}
+                onPageChange={this.onPageChange}
+                items={documentRequests}
+              />
+              <PatronShowLink
+                items={documentRequests}
+                onShowAllClick={this.onShowAll}
+                onShowLessClick={this.onShowLess}
+                rowsPerPage={rowsPerPage}
+              />
               <ResultsTable
                 unstackable
                 data={documentRequests.hits}
@@ -278,18 +299,15 @@ class PatronCurrentDocumentRequests extends Component {
                 name="requests for new literature"
                 currentPage={activePage}
                 renderEmptyResultsElement={this.renderNoResults}
+                showAllResults
               />
-              {documentRequests.total > 0 && (
-                <Container textAlign="center">
-                  <Pagination
-                    currentPage={activePage}
-                    currentSize={rowsPerPage}
-                    loading={isLoading}
-                    onPageChange={this.onPageChange}
-                    totalResults={documentRequests.total}
-                  />
-                </Container>
-              )}
+              <PatronPagination
+                currentPage={activePage}
+                currentSize={rowsPerPage}
+                loading={isLoading}
+                onPageChange={this.onPageChange}
+                items={documentRequests}
+              />
             </Error>
           </ILSItemPlaceholder>
         </>
--- a/src/lib/pages/frontsite/PatronProfile/PatronPastLoans/PatronPastLoans.js
+++ b/src/lib/pages/frontsite/PatronProfile/PatronPastLoans/PatronPastLoans.js
@@ -82,11 +82,12 @@ export default class PatronPastLoans extends Component {
   render() {
     const { error, isLoading, loans, rowsPerPage } = this.props;
     const { activePage } = this.state;
+    const headerTitle = `Your past loans (${loans.total})`;
     return (
       <Container className="spaced">
         <Header
           as="h2"
-          content="Your past loans"
+          content={headerTitle}
           className="highlight"
           textAlign="center"
         />
--- a/src/lib/pages/frontsite/PatronProfile/PatronShowLink.js
+++ b/src/lib/pages/frontsite/PatronProfile/PatronShowLink.js
@@ -1,0 +1,44 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Grid, Icon, Divider } from 'semantic-ui-react';
+
+export const PatronShowLink = ({
+  items,
+  onShowAllClick,
+  onShowLessClick,
+  rowsPerPage,
+}) => {
+  const allHits = items.total === items.hits.length ? true : false;
+  return (
+    items.total > 0 &&
+    items.total > rowsPerPage && (
+      <Grid>
+        <Grid.Column textAlign="center">
+          {!allHits ? (
+            <>
+              <Divider hidden />
+              <div className="link-like" onClick={() => onShowAllClick()}>
+                SHOW ALL <Icon name="caret right" />
+              </div>
+            </>
+          ) : (
+            <div className="link-like" onClick={() => onShowLessClick()}>
+              SHOW LESS <Icon name="caret right" />
+            </div>
+          )}
+        </Grid.Column>
+      </Grid>
+    )
+  );
+};
+
+PatronShowLink.propTypes = {
+  items: PropTypes.object.isRequired,
+  onShowAllClick: PropTypes.func.isRequired,
+  onShowLessClick: PropTypes.func.isRequired,
+  rowsPerPage: PropTypes.number,
+};
+
+PatronShowLink.defaultProps = {
+  rowsPerPage: 5,
+};
--- a/src/lib/pages/frontsite/PatronProfile/PatronShowLink.js
+++ b/src/lib/pages/frontsite/PatronProfile/PatronShowLink.js
@@ -8,7 +8,7 @@ export const PatronShowLink = ({
   onShowLessClick,
   rowsPerPage,
 }) => {
-  const allHits = items.total === items.hits.length ? true : false;
+  const allHits = items.total === items.hits.length;
   return (
     items.total > 0 &&
     items.total > rowsPerPage && (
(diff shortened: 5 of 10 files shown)


jrcastro2:
For the current view I would add a bit more bottom margin/padding, so I would vote for the second one from the "Other potential views".

src/lib/config/defaultConfig.js
```diff
@@ -34,7 +34,7 @@ export const APP_CONFIG = {
   SUCCESS_AUTO_DISMISS_SECONDS: 10,
   DEFAULT_CURRENCY: DEFAULT_CURRENCY,
   DEFAULT_LANGUAGE: DEFAULT_LANGUAGE,
-  DEFAULT_RESULTS_SIZE: 15,
+  DEFAULT_RESULTS_SIZE: 100,
```
kprzerwa, line 37: did you check what influence will this change have on the application?

src/semantic-ui/site/elements/container.overrides
```diff
@@ -268,3 +268,16 @@ Container Overrides - REACT-INVENIO-APP-ILS
     }
   }
 }
+
+.ui.container.link-like {
```
kprzerwa, line 272: why did you create this dependance on the container for your styling? are there already existing classes you could use?

src/semantic-ui/site/elements/container.overrides
```diff
@@ -268,3 +268,16 @@ Container Overrides - REACT-INVENIO-APP-ILS
...
+    color: #4183C4 !important;
```
kprzerwa, line 273: if you hardcode the color like this, nobody will be able to override it in the overlay application (f.e. cds-ils)

src/lib/pages/frontsite/PatronProfile/PatronCurrentDocumentRequests/PatronCurrentDocumentRequests.js
```diff
@@ -233,9 +227,34 @@ class PatronCurrentDocumentRequests extends Component {
...
+        onPageChange={this.onPageChange}
+        items={documentRequests}
+      />
+    );
+  };
+
   render() {
     const { documentRequests, isLoading, error, rowsPerPage } = this.props;
     const { activePage, isSuccessMessageVisible, successMessage } = this.state;
+    const content = `Your requests for new literature (${documentRequests.hits.length}/${documentRequests.total})`;
```
kprzerwa, line 257: I would keep only total number `(16)` as discussed on the chat

src/lib/pages/frontsite/PatronProfile/LoansList.js
```diff
@@ -47,4 +49,5 @@ LoansList.propTypes = {
   rowsPerPage: PropTypes.number.isRequired,
   renderListEntry: PropTypes.func.isRequired,
   noLoansCmp: PropTypes.element.isRequired,
+  renderShow: PropTypes.element,
```
kprzerwa, line 52: what is the reason for passing the component  as a prop? I am asking since I don't remember this part

src/lib/pages/frontsite/PatronProfile/PatronCurrentDocumentRequests/PatronCurrentDocumentRequests.js
```diff
@@ -269,6 +288,13 @@ class PatronCurrentDocumentRequests extends Component {
           )}
           <ILSItemPlaceholder fluid isLoading={isLoading}>
             <Error error={error}>
+              {documentRequests.hits.length > 2 && this.renderPagination()}
```
kprzerwa, line 291: not sure why if there is more than 2 hits then the pagination is showed ? why specifically 2 ?

src/lib/pages/frontsite/PatronProfile/PatronPastLoans/PatronPastLoans.js
```diff
@@ -82,11 +82,12 @@ export default class PatronPastLoans extends Component {
   render() {
     const { error, isLoading, loans, rowsPerPage } = this.props;
     const { activePage } = this.state;
+    const content = `Your past loans (${loans.hits.length}/${loans.total})`;
```
kprzerwa, line 85:
```suggestion
    const headerTitle = `Your past loans (${loans.hits.length}/${loans.total})`;
```

src/lib/pages/frontsite/PatronProfile/PatronShow.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Container, Grid, Icon } from 'semantic-ui-react';
+
+export const RenderShow = ({ items, onShowAll, onShowLess, rowsPerPage }) => {
```
kprzerwa, line 5:
```suggestion
export const ShowAllLink = ({ items, onShowAll, onShowLess, rowsPerPage }) => {
```

src/semantic-ui/site/elements/container.overrides
```diff
@@ -268,3 +268,16 @@ Container Overrides - REACT-INVENIO-APP-ILS
...
+    cursor: pointer !important;
+}
+
+.ui.container.top-spaced {
```
kprzerwa, line 277: did you check how was the space created for the View ALL link on the main page? We should avoid the custom CSS if possible

src/semantic-ui/site/elements/container.overrides
Pineirin: The idea of the show All is to have the appearance of the Link from react-router-dom. Using link is a bad idea, as even if I use as target the page itself (#) the view will move to the top of the page, instead of the top of the element we just refreshed.

src/lib/config/defaultConfig.js
Pineirin: Yes, it affects only the patron profile, which is exactly what we want.

src/lib/pages/frontsite/PatronProfile/PatronCurrentDocumentRequests/PatronCurrentDocumentRequests.js
Pineirin: I saw there was an error in the last push (the last push was only that change). As you can see in the screenshots the change was already implemented.

src/lib/pages/frontsite/PatronProfile/PatronCurrentDocumentRequests/PatronCurrentDocumentRequests.js
Pineirin: Having both pagination at the top and at the bottom makes it look ugly when there are less than three items in the list (both pagination elements are far to close to each other). I decided to not show the top one with two items or less.

src/lib/pages/frontsite/PatronProfile/LoansList.js
Pineirin: This way we can send an element that render the ShowAll button. We need to have this element in the LoansList, as it's placed below the pagination but before the list; but it's also better to create it in the parent elements (PatronCurrentLoans and PatronPendingLoans), so we can use the onShowAll and onShowLess functions to create it there instead of having to pass both function to the child component.

src/semantic-ui/site/elements/container.overrides
Pineirin: It doesn't have a margin (the margin is due to the title it has on top).

src/lib/config/defaultConfig.js
kprzerwa: are you sure? how come all the snapshosts below were modified? You changed the global config, I would expect it affects the application globally

ntarocco: I would suggest to check again.
On top of that, the var is called `DEFAULT_RESULTS_SIZE`: you don't want to change a `default` value, normally globally used, for a few specific components.

src/lib/pages/frontsite/PatronProfile/LoansList.js
```diff
@@ -1,37 +1,39 @@
...
-      loans,
-      rowsPerPage,
-      renderListEntry,
-      noLoansCmp,
-    } = this.props;
+    const { loans, renderListEntry, noLoansCmp, showAllLink } = this.props;
+
     return loans.total > 0 ? (
       <>
+        {loans.hits.length > 2 && this.renderPagination()}
```
ntarocco, line 27: Why are you checking for more than `2`? BTW, you should use `loans.hits.total` instead of computing it (pay attention that it is different in ES6 and ES7, see examples in the code)

src/lib/pages/frontsite/PatronProfile/PaginationShow.js
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import { Pagination } from '@components/Pagination';
+import PropTypes from 'prop-types';
+import { Container } from 'semantic-ui-react';
+
+export const RenderPagination = ({
```
ntarocco, line 6: I would suggest to name components with `names`, not `actions`?
```suggestion
export const Pagination = ({
```

The component name does not match the filename. We normally keep them aligned.
Do you have to create a new cmp?

src/lib/pages/frontsite/PatronProfile/PatronCurrentDocumentRequests/PatronCurrentDocumentRequests.js
ntarocco: See my comment to avoid to use `.length`.
I think there is an issue with the logic of your pagination cmp then: it should not render if only one page.

src/lib/config/defaultConfig.js
Pineirin: It seems I in fact did some mistake when fetching for uses (maybe I bounded the search or something like that). I have corrected this problem using a new configuration variable.

src/lib/pages/frontsite/PatronProfile/PatronShow.js
```diff
@@ -0,0 +1,39 @@
...
+          )}
+        </Grid.Column>
+      </Grid>
+    )
+  );
+};
+
+ShowAllLink.propTypes = {
+  items: PropTypes.object.isRequired,
+  onShowAll: PropTypes.func.isRequired,
```
ntarocco, line 32: Very minor:
```suggestion
  onShowAllClick: PropTypes.func.isRequired,
```
same below

src/lib/config/defaultConfig.js
```diff
@@ -86,6 +86,7 @@ export const APP_CONFIG = {
   },
   HOME_SEARCH_BAR_PLACEHOLDER:
     'Search for books, ebooks, series, journals and standards.',
+  PATRON_MAX_ROWS: 100,
```
ntarocco, line 89: Naming police, sorry: :)
```suggestion
  PATRON_PROFILE_MAX_RESULTS_SIZE: 100,
```

src/semantic-ui/site/elements/container.overrides
kprzerwa: I am not sure how this is influencing the fact the color is hardcoded. You should be able to read it from one of the available LESS variables

src/semantic-ui/site/elements/container.overrides
kprzerwa: yes because the negative space is there created by `<Divider>` not CSS class. That is why I suggested for you too look how it is done there.

src/semantic-ui/site/elements/container.overrides
kprzerwa: I am not sure if it answers my question. Why is this change placed in the `.ui.container`? The style of this component should not be depending on where it is placed (in this case - the container)

src/lib/pages/frontsite/PatronProfile/LoansList.js
```diff
@@ -1,37 +1,39 @@
...
-        <Container textAlign="center">
-          <Pagination
-            currentPage={activePage}
-            currentSize={rowsPerPage}
-            loading={isLoading}
-            onPageChange={onPageChange}
-            totalResults={loans.total}
-          />
-        </Container>
+        {this.renderPagination()}
```
kprzerwa, line 36: is there any reason of creating a new method for this JSX code?

src/lib/pages/frontsite/PatronProfile/PatronCurrentDocumentRequests/PatronCurrentDocumentRequests.js
Pineirin: It doesn't.
The problem this condition is solving is not if there is only page but when there are multiple pages and one of them (the last one) as to few rows.

src/lib/pages/frontsite/PatronProfile/LoansList.js
kprzerwa: I would say in this case is not code duplication since there is no "logic" but only component usage 
```
      <RenderPagination
        currentPage={activePage}
        currentSize={rowsPerPage}
        loading={isLoading}
        onPageChange={onPageChange}
        items={loans}
      />
```

code duplication would be more if you would perform some actions repeatedly

src/lib/pages/frontsite/PatronProfile/PatronCurrentDocumentRequests/PatronCurrentDocumentRequests.js
kprzerwa: so if you are on the last page, and the last page has only one result, then the pagination disappears ? do I understand correctly? If yes, then I would prefer to leave both paginations, since it will prevent the UI from jumping and suddenly changing many things. I think the actionable buttons, as pagination should not change so suddenly, since it can be confusing for the user, who might not understand why suddenly one of the paginations disappeared

src/lib/pages/frontsite/PatronProfile/PatronCurrentDocumentRequests/PatronCurrentDocumentRequests.js
```diff
@@ -233,9 +228,37 @@ class PatronCurrentDocumentRequests extends Component {
     this.setState({ isSuccessMessageVisible: false, successMessage: '' });
   };
 
+  onShowAll = () => {
+    const { patronPid, fetchPatronDocumentRequests } = this.props;
+    fetchPatronDocumentRequests(patronPid, {
```
kprzerwa, line 233: why do we need to pass patronPid here, but not below?

src/lib/pages/frontsite/PatronProfile/PatronCurrentDocumentRequests/PatronCurrentDocumentRequests.js
Pineirin: It looks a bit weird I have to say. I will share a screenshot after pushing again.

src/lib/pages/frontsite/PatronProfile/PatronCurrentDocumentRequests/PatronCurrentDocumentRequests.js
Pineirin: Is not the same function call. This first one calls the function that was passed as prop to the component.
The second calls a function of the component that also calls the passed one with the appropriate parameters.

src/lib/pages/frontsite/PatronProfile/PatronShowLink.js
```diff
@@ -0,0 +1,44 @@
...
+  const allHits = items.total === items.hits.length ? true : false;
+  return (
+    items.total > 0 &&
+    items.total > rowsPerPage && (
+      <Grid>
+        <Grid.Column textAlign="center">
+          {!allHits ? (
+            <>
+              <Divider hidden />
+              <Container className="link-like" onClick={() => onShowAllClick()}>
```
kprzerwa, line 20:
```suggestion
              <a href="#" onClick={() => onShowAllClick()}>
```
you will probably avoid adding custom CSS in this way (to be checked)

src/semantic-ui/site/globals/site.variables
```diff
@@ -28,6 +28,8 @@
 @headlineBackground       : url("@{imagesFolder}/home-headline-background.jpg");
 @loginBackground          : url("@{imagesFolder}/login-background.png");
 @boSidebarWidth           : 230px;
+@linkColor                : #4183C4;
```
kprzerwa, line 31: this variable is already provided by the semantic-ui theme, so you are re-declaring it here

src/lib/pages/frontsite/PatronProfile/PatronShowLink.js
Pineirin: The drawback of this implementation (tested) was that it sends you to the top of the page (instead of the top of the refreshed component).

src/lib/components/ResultsTable/ResultsTable.js
```diff
@@ -26,6 +26,8 @@ export class ResultsTable extends Component {
       ...tableProps
     } = this.props;
 
+    console.log(data);
```
kprzerwa, line 29: 👀

src/lib/pages/frontsite/PatronProfile/LoansList.js
```diff
@@ -1,37 +1,45 @@
...
     return loans.total > 0 ? (
       <>
+        <PatronPagination
+          currentPage={activePage}
+          currentSize={rowsPerPage}
+          loading={isLoading}
+          onPageChange={onPageChange}
+          items={loans}
+        />
+        {patronShowLink && patronShowLink}
```
kprzerwa, line 28: probably should work properly with only
```suggestion
        {patronShowLink}
```

src/lib/pages/frontsite/PatronProfile/PatronShowLink.js
kprzerwa: thats probably the fault of `href="#" ` did you try other options? For example `href={javascript:void(0)}`?

Pineirin: I have tested using javascript:void(0) in an "a" tag and it works as intended. Still, in the case of the Link element, I didn't manage to make it work as we want. 
I'm reluctant to use an "a" with a void href here instead of the current implementation. I see the point of trying to use Link, both to keep the use of elements consistent with the rest of the applications and to avoid overriding css. But I just don't see the point of using "a" here, because: we are not being consistent, we still need to use custom css to keep it as in the links of the frontpage, and we are using an hyperlink tag to something whose functionality is way more closer to a button than to a link (in fact, it doesn't involve any navigation at all).

ntarocco: The only advantage of hyperlinks, at least the one I can think of, is accessibility: screen readers will "read" the cmp as a link: in this case, a `Container` is probably a `<div>` and a screen reader will probably not treat it as clickable.
Given said so, we did not put efforts in making InvenioILS accessible, but I would say that we can make efforts to add something that is not accessible.
Did you have a look [here](https://stackoverflow.com/questions/1631924/preventing-onclick-page-jumps), for example?

src/lib/pages/frontsite/PatronProfile/PatronShowLink.js
```diff
@@ -0,0 +1,44 @@
...
+  const allHits = items.total === items.hits.length ? true : false;
+  return (
+    items.total > 0 &&
+    items.total > rowsPerPage && (
+      <Grid>
+        <Grid.Column textAlign="center">
+          {!allHits ? (
+            <>
+              <Divider hidden />
+              <div className="link-like" onClick={() => onShowAllClick()}>
```
ntarocco, line 20: No link in the end? Did you see my comment about `accessbility`?

kprzerwa: I understood from your comment that we don't want the <a> tag there because of the screen readers, but now I think I misunderstood

PR #519, opened 2022-01-06
Patron: auto-checkout on paste

NRodriguezcuellar:
Now automatically checks out the found item if a barcode is pasted in the checkout section of the patron details page.

closes https://github.com/CERNDocumentServer/cds-ils/issues/546

### DIFF
--- a/src/lib/pages/backoffice/Patron/PatronDetails/ItemsSearch/ItemsSearch.js
+++ b/src/lib/pages/backoffice/Patron/PatronDetails/ItemsSearch/ItemsSearch.js
@@ -1,4 +1,3 @@
-import { recordToPidType } from '@api/utils';
 import { Error } from '@components/Error';
 import { Loader } from '@components/Loader';
 import { SearchBarILS } from '@components/SearchBar';
@@ -24,39 +23,27 @@ export default class ItemsSearch extends Component {
     clearResults();
   };
 
-  executeSearch = (queryString) => {
-    const { queryString: propsQueryString, fetchItems } = this.props;
+  executeSearch = async (queryString, patronPid) => {
+    const { queryString: propsQueryString, fetchAndCheckoutIfOne } = this.props;
     queryString = queryString || propsQueryString;
+
     // eslint-disable-next-line react/no-unused-state
-    this.setState({ prevSearchQuery: queryString, executedSearch: true });
-    fetchItems(queryString);
+    this.setState({ prevSearchQuery: queryString, executedSearch: true }, () =>
+      fetchAndCheckoutIfOne(queryString, patronPid)
+    );
   };
 
   onPasteHandler = async (event) => {
-    const { checkoutItem, patronDetails } = this.props;
     const { prevSearchQuery } = this.state;
+    const { patronDetails } = this.props;
+
     let queryString = event.clipboardData.getData('Text');
+
     const sameQueryString = prevSearchQuery === queryString;
 
     if (queryString && !sameQueryString) {
-      await this.executeSearch(queryString);
-      const {
-        items: { hits },
-      } = this.props;
+      await this.executeSearch(queryString, patronDetails.user_pid);
 
-      const hasOneHit =
-        !_isEmpty(hits) &&
-        hits.length === 1 &&
-        hits[0].metadata.status === 'CAN_CIRCULATE';
-
-      if (hasOneHit) {
-        const documentPid = hits[0].metadata.document.pid;
-        const itemPid = {
-          type: recordToPidType(hits[0]),
-          value: hits[0].metadata.pid,
-        };
-        await checkoutItem(documentPid, itemPid, patronDetails.user_pid, true);
-      }
       // eslint-disable-next-line react/no-unused-state
       this.setState({ prevSearchQuery: '', executedSearch: true });
     }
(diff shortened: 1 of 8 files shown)


src/lib/pages/backoffice/Patron/PatronDetails/ItemsSearch/index.js
```diff
@@ -4,7 +4,8 @@ import ItemsSearchComponent from './ItemsSearch';
 import { clearResults, fetchItems, updateQueryString } from './state/actions';
 
 const mapDispatchToProps = (dispatch) => ({
-  fetchItems: (barcode) => dispatch(fetchItems(barcode)),
+  fetchItems: (barcode, autoCheckout, patron) =>
+    dispatch(fetchItems(barcode, autoCheckout, patron)),
```
ntarocco, line 8: Isn't it expecting the `pid` of the patron and not the patronDetails obj (even if I am not sure what is inside)?
```suggestion
    dispatch(fetchItems(barcode, autoCheckout, patronPid)),
```

src/lib/pages/backoffice/Patron/PatronDetails/ItemsSearch/state/actions.js
```diff
@@ -22,6 +25,26 @@ export const fetchItems = (barcode) => {
         type: SUCCESS,
         payload: response.data,
       });
+
+      if (autoCheckout && patron) {
```
ntarocco, line 29: It looks like that the feature was already implemented, but maybe not working?

We should not have a `fetchItems` func to have extra responsibilities, in particular to perform the action of checking out.
Can it stay where it was? What was the issue?

src/lib/pages/backoffice/Patron/PatronDetails/ItemsSearch/index.js
NRodriguezcuellar: ah yes, you're right

src/lib/pages/backoffice/Patron/PatronDetails/ItemsSearch/state/actions.js
NRodriguezcuellar: yes that is right, there was an implementation but it could can't work because the prop does not get updated in time.

NRodriguezcuellar: i seems to work with a small timeOut :sweat_smile:

src/lib/pages/backoffice/Patron/PatronDetails/ItemsSearch/ItemsSearch.js
```diff
@@ -24,39 +24,56 @@ export default class ItemsSearch extends Component {
...
+        const {
+          items: { hits },
+        } = this.props;
+
+        const hasOneHit = this.hitAvailableForCheckout(hits);
+
+        if (hasOneHit) {
+          await this.checkoutItem(hits[0]);
+        }
+      }, 600);
```
ntarocco, line 75: We cannot put a timeout because, in this case, it highly depends on the time that the HTTP request will take to fetch the data. We cannot control it.
In general, a timeout is, in my opinion, a code smell.

After having checked the implementation of the original React component, it is clear to me that it has not been implemented in the correct way.
What you have done in your previous implementation was more correct: my comment was referring to the fact that it is probably not ideal to add a `checkout` instead of a `fetchItems` method.

Suggestion to fix it. In Redux action, check if [fetchItems](https://github.com/inveniosoftware/react-invenio-app-ils/blob/master/src/lib/pages/backoffice/Patron/PatronDetails/ItemsSearch/state/actions.js#L10) is called/reused somewhere else.

## if already called

- create a new action e.g. `fetchAndCheckoutIfOne` (please choose a better name :)) that will call the other `fetchItems` and then perform the checkout (move the checkout code to the action)

## if not already called

- as your previous implementation, move the checkout inside and then rename the method, so that you are sure that on other component in the future will use that function to just fetch items (and will have as side effect to checkout).

src/lib/pages/backoffice/Patron/PatronDetails/ItemsSearch/state/const.js
```diff
@@ -0,0 +1,6 @@
+//separated into different file because of a circular dependency caused otherwise
```
kprzerwa, line 1: could you rename the file to `states.js` ? I think it describes better the nature of these constants

kprzerwa (commented):
LGTM, can we have a demo ? :)

src/lib/pages/backoffice/Patron/PatronDetails/ItemsSearch/state/const.js
NRodriguezcuellar: Maybe constants or types? I think states is also a bit misleading? As it does not contain any state just the action types. What do you think?

kprzerwa: good idea with `types.js`, this is how we called these files before we decided to merge them with actions.

NRodriguezcuellar: Will go with that then 👍

ntarocco (approved):

src/lib/pages/backoffice/Patron/PatronDetails/ItemsSearch/state/types.js
```diff
@@ -0,0 +1,6 @@
+//separated into different file because of a circular dependency caused otherwise
+export const IS_LOADING = 'itemsSearchBarcode/IS_LOADING';
```
ntarocco, line 2: In the past, we have removed all the `types.js` and moved the consts inside `action.js` to avoid yet another file, just for that.
I would keep as it is for consistency, but this is very very minor.

PR #526, opened 2022-01-31
backoffice: Add alternative titles to details page

jrcastro2:
## ALternative titles
image

## Alternative abstracts
image

### DIFF
--- a/src/lib/modules/Document/DocumentAlternativeAbstracts.js
+++ b/src/lib/modules/Document/DocumentAlternativeAbstracts.js
@@ -1,0 +1,36 @@
+import PropTypes from 'prop-types';
+import React, { Component } from 'react';
+import Overridable from 'react-overridable';
+import { List } from 'semantic-ui-react';
+import _isEmpty from 'lodash/isEmpty';
+import { ShowMoreContent } from '@components/ShowMoreContent';
+
+class DocumentAlternativeAbstracts extends Component {
+  render() {
+    const {
+      metadata: { alternative_abstracts: alternativeAbstracts },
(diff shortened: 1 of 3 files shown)


kprzerwa:
Did you verify where the alternative titles were before the latest changes erased them? In the backoffice the type of the subtitle should be also displayed

kprzerwa:
> > Did you verify where the alternative titles were before the latest changes erased them? In the backoffice the type of the subtitle should be also displayed
> 
> Yes they were removed in [this PR](https://github.com/inveniosoftware/react-invenio-app-ils/commit/edbadc946e3707df25018e3e41527aaaa3e00798#diff-59302d0af55bacb5efd5b0150da8e906f8cfe26c27c2516461a07595c91ec7b9L37-L44) apparently there was an "others" tab. I displayed the information as it is done in the frontsite. Note that the alternative_abstract is also not displayed in the backoffice, should we also add it in the content tab?
> 
> In order to display the type of the alternative title I would propose to move it down, as it is done for the identifiers (see image below), WDYT? image

in general I think we should think here what is the best for the librarians UX. We have the choice between having alternative titles next to the title (convenience) and not giving the secondary information (i.e alternative titles) so much meaning to move it up into the prominent place in the metadata display, additionally, the information takes space in the UI.
Personally I wouldn't give them a separate section as the identifiers have, for the simple reason that they are not as important. It it not as common to have this information either, so I would leave it next to the title as you did, but displayed as follows:
```
- alter title (ALTERNATIVE TITLE TYPE)
- ...
```

kprzerwa (approved):

Issue #424, opened 2021-03-31
Cover doesn't appear in attach multipart modal


KonstantinaStoikou:
Package version (if known): v1.0.0-alpha.34

## Describe the bug
The cover doesn't appear in the `Attach multipart` modal

## Steps to Reproduce
1. Go to the document details in the backoffice
2. Click on `Attach multipart` or `Add to a serial`
3. Pick a multipart/serial
4. The cover is not appearing

## Expected behavior
It should display the cover when a multipart/serial is selected

## Screenshots (if applicable)
![Screenshot 2021-03-31 at 09 54 13](https://user-images.githubusercontent.com/33685068/113110865-dcb1d380-9207-11eb-9f47-4669c3cf6374.png)

PR #545, opened 2022-07-27
attach multipart modal: fix cover image appearing

anikachurilova:
* add missing covering component to multipart and serials attaching modals
* fix overridable component name
* closes https://github.com/inveniosoftware/react-invenio-app-ils/issues/424

Covers appearing preview:
cover appearing prefill

### DIFF
--- a/src/lib/modules/Relations/backoffice/components/RelationCard/SeriesCard.js
+++ b/src/lib/modules/Relations/backoffice/components/RelationCard/SeriesCard.js
@@ -1,4 +1,3 @@
-import { SeriesIcon } from '@components/backoffice/icons';
 import LiteratureTitle from '@modules/Literature/LiteratureTitle';
 import { SeriesAuthors } from '@modules/Series/SeriesAuthors';
 import { BackOfficeRoutes } from '@routes/urls';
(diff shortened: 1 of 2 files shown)


src/lib/modules/Relations/backoffice/components/RelationCard/SeriesCard.js
```diff
@@ -18,7 +21,14 @@ export default class SeriesCard extends Component {
           {actions}
           {data.metadata.series_type || data.metadata.mode_of_issuance}
         </Card.Meta>
-        <SeriesIcon size="huge" color="grey" />
+        {recordToPidType(data) === 'serid' ? (
```
NRodriguezcuellar, line 24: Suggestion: variables!

Honestly it is pretty readable already but i just think it is always a good practice to save conditions into variables, they allow you to actually describe what you are evaluating!  😃 

```js
// or some other name if this is not the thing you are evaluating
const isSerialRecord = recordToPidType(data) === 'serid' 

isSerialRecord? <> : <>

```

_also tell me how you make the screen captures_

NRodriguezcuellar (approved):

src/lib/modules/Relations/backoffice/components/RelationCard/SeriesCard.js
kpsherva: generally I agree with Nicolases suggestion.
I would go step further though and try to understand why you need to check if the record is series? It is already specific `SeriesCard` so it should only be used for series, what was the need for this additional check?

kpsherva (approved):

PR #547, opened 2022-08-10
http error pages: add page for 400 error code

anikachurilova:
* add component for displaying bad request error page
* fix 400-error related tests
* addresses https://github.com/CERNDocumentServer/cds-ils/issues/759

Bad Request error page preview:
Screenshot 2022-08-08 at 14 52 33

### DIFF
--- a/src/lib/api/base.js
+++ b/src/lib/api/base.js
@@ -11,7 +11,7 @@ const apiConfig = {
 };
 const http = axios.create(apiConfig);
 
-const HTTP_STATUS_CODES_WITH_ERROR_PAGE = [404, 429, 500];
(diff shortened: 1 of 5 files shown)


src/lib/api/base.js
```diff
@@ -11,7 +11,7 @@ const apiConfig = {
 };
 const http = axios.create(apiConfig);
 
-const HTTP_STATUS_CODES_WITH_ERROR_PAGE = [404, 429, 500];
+const HTTP_STATUS_CODES_WITH_ERROR_PAGE = [400, 404, 429, 500];
```
ntarocco, line 14: I am not 100% sure about this: if I am not mistaken, when there is a 400 error from the backend, it shows a banner on the top right.
To test it, you should try, for example, to check-out a loan that is in "completed" state and check the response code.

PR #559, opened 2022-10-24
V1.0.0 alpha.72

zzacharo:


kpsherva (approved):

src/lib/forms/rjsf/widgets/RJSFVocabulary.js
```diff
@@ -133,6 +136,7 @@ RJSFVocabulary.propTypes = {
   value: PropTypes.string,
   options: PropTypes.shape({
     vocabularyType: PropTypes.string.isRequired,
+    size: PropTypes.number,
```
kpsherva, line 139: missing default ?

zzacharo: discuss IRL: options is an object so the size will not be defaulted as we pass as `options: {vocabularyType: '...'}`.

PR #580, opened 2023-12-12
backoffice: patron: checkout: Fix for Search/Scan Barcode and auto checkout loan if one

sakshamarora1:

### Description

In the checkout tab, the barcode scanner didn't auto-checkout a book on being scanned for a patron due to `onSearchHandler` not passing the patronPid argument to the executeSearch function. This PR change will use the patronDetails available in `this.props` and will fill in the patronPid whenever the parameter is undefined

Fixes: https://github.com/CERNDocumentServer/cds-ils/issues/546

Note:
The barcode scanner does the following things on a scan -
-> Type in the barcode
-> Press the return key

### Checklist

Ticks in all boxes and 🟢 on all GitHub actions status checks are required to merge:

- [x] I'm aware of the [code of conduct](https://inveniordm.docs.cern.ch/contribute/code-of-conduct/).
- [x] I've created [logical separate commits](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commits) and followed the [commit message format](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commit-message).
- [x] I've followed the [CSS/JS](https://inveniordm.docs.cern.ch/develop/best-practices/css-js/) and [React](https://inveniordm.docs.cern.ch/develop/best-practices/react/) guidelines (for relevant code).
- [x] I've followed the [web accessibility](https://inveniordm.docs.cern.ch/develop/best-practices/accessibility/) guidelines (for relevant code).
- [x] I've followed the [user interface](https://inveniordm.docs.cern.ch/develop/best-practices/ui/) guidelines (for relevant code).
- [x] I've NOT included third-party code (copy/pasted source code or new dependencies).

https://github.com/inveniosoftware/react-invenio-app-ils/assets/50872172/ee037db4-131a-4b34-8cfb-bae7fe6af91d

### DIFF
--- a/src/lib/pages/backoffice/Patron/PatronDetails/ItemsSearch/ItemsSearch.js
+++ b/src/lib/pages/backoffice/Patron/PatronDetails/ItemsSearch/ItemsSearch.js
@@ -24,8 +24,13 @@ export default class ItemsSearch extends Component {
   };
 
   executeSearch = async (queryString, patronPid) => {
-    const { queryString: propsQueryString, fetchAndCheckoutIfOne } = this.props;
+    const {
+      queryString: propsQueryString,
+      patronDetails,
+      fetchAndCheckoutIfOne,
+    } = this.props;
     queryString = queryString || propsQueryString;
+    patronPid = patronPid || patronDetails.user_pid;
 
     // eslint-disable-next-line react/no-unused-state
     this.setState({ prevSearchQuery: queryString, executedSearch: true }, () =>


kpsherva (approved):

kpsherva (commented):
🎊 congrats for the first PR :)
well done, the only suggestion is commit message:
```
backoffice: pass patron details on checkout item search
```

src/lib/pages/backoffice/Patron/PatronDetails/ItemsSearch/ItemsSearch.js
```diff
@@ -24,8 +24,13 @@ export default class ItemsSearch extends Component {
   };
 
   executeSearch = async (queryString, patronPid) => {
-    const { queryString: propsQueryString, fetchAndCheckoutIfOne } = this.props;
+    const {
```
zzacharo, line 27: Quickly checking around, as you described, the patronPid is not passed from the search handlers. Can you confirm this and if so, better remove the argument from the `executeSearch` function?

sakshamarora1: The `patronPid` is passed from the `onPasteHandler` function to the `executeSearch` function, but it can be removed and then the `patronDetails` will always be fetched from the props. I've tested this new change and it works the same, should I make this update?

zzacharo: I would say so, so we know that information comes from one source only. @kpsherva wdyt?

ntarocco: I also don't understand this change well. The `executeSearch` is called as the following:
`await this.executeSearch(queryString, patronDetails.user_pid);`
Why using it via props instead?

From a quick check, the issue looks related to the `onSearchHandler`, which does not pass the `patronPid`. If that is the case, then yes, the signature of the func must be changed and the `patronPid` param should be removed because it will be always `null` onSearch, while it will be correctly set `onPaste`.

kpsherva: I would prefer not to risk breaking compatibility now, there are places where the arguments are passed in non explicit way so I prefer not to change the signature

ntarocco: The signature is already broken ;) I suspect that when you search, the param `patronPid` is `null`.

Issue #581, opened 2023-12-12
backoffice: check-out: App crashes when physical copy is scanned before scanning the patron ID


sakshamarora1:
Package version: [v1.0.0-alpha.77](https://github.com/inveniosoftware/react-invenio-app-ils/releases/tag/v1.0.0-alpha.77)

## Describe the bug
When checking out a physical copy by scanning the physical copy first and then scanning the patron ID, the application creates multiple loans for the same physical copy and then gives an internal server error, crashing the app.

## Steps to Reproduce
1. Go to 'Check-out' tab
2. Scan the Physical Copy's barcode first 
3. Scan the patron ID on the 'You are about to checkout the physical copy with barcode' popup dialog box

## Expected behaviour

Backend: It should only create a single loan.
Frontend: It should only send 1 checkout request to backend.

## Screenshots (if applicable)

https://github.com/inveniosoftware/react-invenio-app-ils/assets/50872172/a666ce9d-2e71-4672-bdcc-eccab7b04f1a

PR #582, opened 2023-12-18
modules: HitsSearch: Fixed debounce search function to not send requests to backend on every keystroke

sakshamarora1:

### Description

Fixes: #581 
- Fixed debounce search function by not re-defining it in every re-render and not send request to backend on every re-render.
- Updated Functionality to only auto-checkout when only 1 result is found and enter is pressed.

### DIFF
--- a/src/lib/modules/ESSelector/HitsSearch.js
+++ b/src/lib/modules/ESSelector/HitsSearch.js
@@ -65,52 +65,46 @@ export class HitsSearch extends Component {
     this.searchInputRef.focus();
   };
 
-  search = (searchQuery) => {
+  debouncedSearch = debounce(async (searchQuery) => {
     const {
       serializer,
       alwaysWildcard,
-      onResults,
-      delay,
       query: queryFunc,
+      onResults,
     } = this.props;
+    let results, hasError;
+    try {
+      const queryString = alwaysWildcard ? searchQuery + '*' : searchQuery;
+      const response = await queryFunc(queryString);
+      results = [];
 
-    const deb = debounce(async (searchQuery) => {
-      let results, hasError;
-      try {
-        const queryString = alwaysWildcard ? searchQuery + '*' : searchQuery;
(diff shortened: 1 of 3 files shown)


src/lib/modules/ESSelector/HitsSearch.js
```diff
@@ -162,6 +158,13 @@ export class HitsSearch extends Component {
     );
   };
 
+  handleKeyPress = async (event) => {
+    const { onResults } = this.props;
+    if (event.key === 'Enter' && onResults !== undefined) {
+      this.setState({ execute: true });
```
ntarocco, line 164: Was this solution of adding a new `state` discussed? Can we chat about this IRL?

ntarocco (commented):
If you have time and they are easy, don't hesitate to also slowly fix some of the linter warning issues in the PR ;)

src/lib/modules/ESSelector/HitsSearch.js
ntarocco: As discussed IRL, explore the option of bubbling up the `onKeyPress` event, and implement the login of checking-out when clicking `Enter` in the `ItemActionMenu`: this would move the extra new state up, closer to the feature where it is needed.

src/lib/pages/backoffice/Item/ItemDetails/ItemActionMenu/ItemActionMenu.js
```diff
@@ -136,8 +170,10 @@ export default class ItemActionMenu extends Component {
           autoSelect={hasCheckOutItem}
           minCharacters={hasCheckOutItem ? 1 : 3}
           onResults={this.onResults}
+          handleKeyPress={this.handleKeyPress}
+          onSearchChange={this.onSearchChange}
           trigger={this.checkoutItemButton()}
-          query={patronApi.list}
+          query={this.patronSearch}
```
kpsherva, line 176: 👍

PR #585, opened 2024-01-12
components: Notifications: Configure error notification to not be added for http code 410

sakshamarora1:

### Description

Fixes: https://github.com/CERNDocumentServer/cds-ils/issues/849

Screenshot 2024-01-12 at 15 06 45

### DIFF
--- a/src/lib/api/base.js
+++ b/src/lib/api/base.js
@@ -12,7 +12,7 @@ const apiConfig = {
 const http = axios.create(apiConfig);
 
 // List used to specify whether to redirect to route `/errors` & prevent a redundant error notifation
-const HTTP_STATUS_CODES_WITH_ERROR_PAGE = [404, 429, 500];
+const HTTP_STATUS_CODES_WITH_ERROR_PAGE = [404, 410, 429, 500];
 // The corresponding dedicated error page components exist in `@pages/frontsite/ErrorPafe`
 const URLS_NOT_TO_REDIRECT_IF_UNAUTHORIZED = ['/me', '/me/loans'];
 // CSRF possible errors
--- a/src/lib/api/base.js
+++ b/src/lib/api/base.js
@@ -11,7 +11,9 @@ const apiConfig = {
 };
 const http = axios.create(apiConfig);
 
+// List used to specify whether to redirect to route `/errors` & prevent a redundant error notifation
 const HTTP_STATUS_CODES_WITH_ERROR_PAGE = [404, 429, 500];
+// The corresponding dedicated error page components exist in `@pages/frontsite/ErrorPafe`
 const URLS_NOT_TO_REDIRECT_IF_UNAUTHORIZED = ['/me', '/me/loans'];
 // CSRF possible errors
 const CSRF_ERROR_REASON_NO_COOKIE = 'CSRF cookie';
@@ -82,4 +84,9 @@ const responseRejectInterceptor = (error) => {
 
 http.interceptors.response.use(undefined, responseRejectInterceptor);
 
-export { http, apiConfig, responseRejectInterceptor };
+export {
+  http,
+  apiConfig,
+  responseRejectInterceptor,
+  HTTP_STATUS_CODES_WITH_ERROR_PAGE,
+};
--- a/src/lib/components/Error/Error.js
+++ b/src/lib/components/Error/Error.js
@@ -4,12 +4,17 @@ import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 import { Container, Message } from 'semantic-ui-react';
 import { DefaultFallbackComponent } from './DefaultFallbackComponent';
+import { HTTP_STATUS_CODES_WITH_ERROR_PAGE } from '@api/base';
 
 export const isAPIError = (error) => {
   return get(error, 'response.data.message') !== undefined;
 };
 
-const shouldShowErrorPage = (error) => {
+export const shouldShowErrorPage = (error) => {
+  if (HTTP_STATUS_CODES_WITH_ERROR_PAGE.includes(error?.response?.status)) {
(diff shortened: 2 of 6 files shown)


src/lib/pages/frontsite/ErrorsPage/ErrorsPage.js
```diff
@@ -17,6 +17,13 @@ export class ErrorsPage extends Component {
         return <InternalServerError />;
       } else if (params.errorCode === 404) {
         return <NotFound />;
+      } else if (params.errorCode === 410) {
+        return (
+          <NotFound
+            title="Resource No Longer Available"
+            message="The requested content has been removed by the library manager"
```
jrcastro2, line 24: suggestion: Just a personal preference.
```suggestion
            message="The requested content has been removed."
```

src/lib/components/Notifications/actions.js
```diff
@@ -1,11 +1,21 @@
...
 export const sendErrorNotification = (error) => {
   return (dispatch) => {
+    const errorStatus = get(error, 'response.data.status');
+    if (
+      errorStatus !== undefined &&
+      HTTP_STATUS_CODES_WITHOUT_NOTIFICATION.includes(errorStatus)
+    ) {
+      dispatch(clearNotifications);
+      return;
+    }
```
jrcastro2, line 18: Minor: Just to confirm I understand correctly the purpose of this code, this is done to avoid showing a notification when redirecting to the 410 page? if that's the case I would have a look to the function `isAPIError` and modifiy the behaviour there rather than adding a custom behaviour here.

src/lib/api/base.js
```diff
@@ -11,7 +11,8 @@ const apiConfig = {
 };
 const http = axios.create(apiConfig);
 
-const HTTP_STATUS_CODES_WITH_ERROR_PAGE = [404, 429, 500];
+const HTTP_STATUS_CODES_WITH_ERROR_PAGE = [404, 410, 429, 500];
+const HTTP_STATUS_CODES_WITHOUT_NOTIFICATION = [410];
```
ntarocco, line 15: I am wondering if all the error codes above should not show a notification, given that they will already redirect to a dedicated error page. WDYT @kpsherva ?

src/lib/components/Notifications/actions.js
sakshamarora1: I feel that the `isAPIError` function should just be a function to specify whether the response was an error or not as it's name suggests.
Adding the notification behaviour there might be making it more complex. What do you think?

sakshamarora1: Alright, thanks for the insight here, I'll update the functions here and reuse them!

src/lib/api/base.js
kpsherva: if I remember correctly, not all the codes have the error page, that was the original bug in the first place - and I agree, the ones which have error page, should not show the notification by default.
In this case the solution would be to implement those pages, instead of adding additional condition (without notification)

sakshamarora1: So the code lists here should be for showing error page and for notifications (and not without as it currently is here) separately?
That way the existing pages won't have the notification anymore and we can configure codes where we just want to show a notification by just adding the code in the notification list.

src/lib/pages/frontsite/ErrorsPage/ErrorsPage.js
```diff
@@ -17,6 +17,13 @@ export class ErrorsPage extends Component {
         return <InternalServerError />;
       } else if (params.errorCode === 404) {
         return <NotFound />;
+      } else if (params.errorCode === 410) {
+        return (
+          <NotFound
```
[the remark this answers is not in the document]
sakshamarora1: Okay, I'll update component name!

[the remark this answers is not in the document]
sakshamarora1: For 5xx, there is the `InternalServerError` component from line 13

src/lib/components/Error/Error.js
```diff
@@ -9,14 +9,13 @@ export const isAPIError = (error) => {
...
-const shouldShowErrorPage = (error) => {
-  if (isAPIError(error)) {
-    // do not display the error in a full page, show the error only in the specific component (probably a notification).
-    return false;
-  }
-  // when the exception is handled, we don't want to display a full page error (probably a notification).
+export const shouldShowErrorPage = (error) => {
   const handledException = error?.response?.status === 400;
-  return handledException ? false : true;
+  if (!handledException || !isAPIError(error)) {
```
ntarocco, line 14: Feel free to keep this like this, the only comment I have is that, in general, to improve readability, we try avoiding double negation. It is more natural to read "positive" conditions.
Here you could:
```javascript
if (handledException || isAPIError(error)) {
  // don't display full page error
  return false; 
} else {
  return true;
}
```

ntarocco (approved):

src/lib/components/HttpErrors/HttpErrors.js
```diff
@@ -60,24 +60,27 @@ ForbiddenComponent.defaultProps = defaultProps;
 
 export const Forbidden = Overridable.component('Forbidden', ForbiddenComponent);
 
-function NotFoundComponent(props) {
+function HttpErrorComponent(props) {
```
ntarocco, line 63: If you want to avoid the renaming, the fix of the imports, and the other PR (pseudocode):
```javascript
function HttpErrorComponent(props) {
...
}


function NotFoundComponent(props) {
  return HttpErrorComponent('Page Not Found', ...)

```

basically, simply create the generic HttpErrorComponent, require title, message, optionally icon. Then, use it in the specialized `NotFoundComponent` passing the right props.

kpsherva (approved):

Issue #591, opened 2024-02-08
AlternativeTitle: App crashes when you add an alternative title without selecting a type for it


sakshamarora1:
Note:
the type field in Alternative Title schema is not mandatory

https://github.com/inveniosoftware/react-invenio-app-ils/assets/50872172/64a55b0a-3b25-4286-83c3-5eb1c0f41f42

PR #592, opened 2024-02-09
AlternativeTitle: Pass default if type key doesn't exist in data object

sakshamarora1:

### Description

Fixes: #591

Screenshot 2024-02-09 at 10 09 51


src/lib/pages/backoffice/Document/DocumentDetails/DocumentMetadata/DocumentMetadataGeneral.js
```diff
@@ -70,9 +70,7 @@ export class DocumentMetadataGeneral extends Component {
           <List bulleted>
             {document.metadata.alternative_titles.map((entry) => (
               <List.Item key={entry.value}>
-                <List.Content>
-                  {entry.value + '(' + entry.type + ')'}
```
kpsherva, line 74: We need to show title type if present, please add it back. especially if it is librarians view, they need all the info displayed

sakshamarora1:
For DocumentDetails in backoffice:
`Test` doesn't have a type, handled by default.

Screenshot 2024-02-09 at 15 48 11

kpsherva (approved):

PR #593, opened 2024-02-13
forms: rjsf: Extend ignoreEnterEvent for dynamically created text fields

sakshamarora1:

### Description

Fixes: https://github.com/CERNDocumentServer/cds-ils/issues/853

### DIFF
--- a/src/lib/forms/rjsf/RJSFCustomTemplates.js
+++ b/src/lib/forms/rjsf/RJSFCustomTemplates.js
@@ -8,6 +8,7 @@ import _find from 'lodash/find';
 import PropTypes from 'prop-types';
 import React from 'react';
 import { Divider, Form, Grid } from 'semantic-ui-react';
+import { ignoreRJSFEnterEvent } from './utils';
 
 /**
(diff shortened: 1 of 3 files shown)


src/lib/forms/rjsf/RJSFCustomTemplates.js
```diff
@@ -236,6 +237,7 @@ ObjectFieldTemplateWrapperGrid.defaultProps = {};
  */
 export function ArrayFieldTemplateWithWrapper(props) {
   const { children, uiSchema } = props;
+  ignoreRJSFEnterEvent(document);
```
jrcastro2, line 240: Comment: Hmm I understand that we need to add this here because at the form cmp level these are not loaded as we can add new fields dynamically.

The only issue I can see is that in terms of performance this is worse as it will loop over items that were already processed. I would rather set the `onkeydown` when creating it, but in a first look it doesn't seem easy to do with rjsf ... However it's probably kicking in only when working on array fields so it should be good enough IMHO.

jrcastro2 (commented):
I would check the performance impact on the form, and if it's little to none it looks good to me :+1:

PR #596, opened 2024-02-27
backoffice: Use Truncate for OrdersList and Patrons Doc Requests table

sakshamarora1:

### Description

Fixes: https://github.com/CERNDocumentServer/cds-ils/issues/821

Screenshot 2024-02-27 at 15 08 33

Screenshot 2024-02-27 at 15 04 48

### DIFF
--- a/src/lib/pages/backoffice/Acquisition/Order/OrderSearch/OrderList/OrderListEntry.js
+++ b/src/lib/pages/backoffice/Acquisition/Order/OrderSearch/OrderList/OrderListEntry.js
@@ -6,6 +6,7 @@ import {
   BackOfficeRoutes,
   ProviderRoutes,
 } from '@routes/urls';
+import { Truncate } from '@components/Truncate';
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 import { Link } from 'react-router-dom';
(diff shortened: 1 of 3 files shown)


src/lib/pages/backoffice/Acquisition/Order/OrderSearch/OrderList/OrderListEntry.js
```diff
@@ -159,18 +160,20 @@ export default class OrderListEntry extends Component {
             to={AcquisitionRoutes.orderDetailsFor(record.metadata.pid)}
             data-test={`navigate-${record.metadata.pid}`}
           >
-            <AcquisitionOrderIcon />
-            Order: {docTitle ? docTitle : record.metadata.pid}
+            <Truncate lines={2} width="500px">
```
jrcastro2, line 163: Comment: Ideally we try to use relative units, however since the backoffice is not meant to be fully responsive, I think it's good enough.

jrcastro2 (approved):
LGTM!

jrcastro2:
Having a look to the screenshots, I would fill in the white space on the right with the title as well, it should not overlap but I feel it's too short like that, WDYT?
image

src/lib/pages/backoffice/Acquisition/Order/OrderSearch/OrderList/OrderListEntry.js
kpsherva: could you check if relative units work? fe 80% ?

sakshamarora1: Yes, I used relative units at first and it worked on 50% but it was not responsive enough if the width was decreased a bit too much.
(My reasoning was that back office wouldn't be used in mobile view)
I went with what we already have in the Documents page i.e. 500px

PR #599, opened 2024-03-08
routes: Render browser tab title based on literature title

sakshamarora1:

### Description

Related: https://github.com/CERNDocumentServer/cds-ils/issues/793

Note: This is just an example PR on how the tab title rendering would look code wise, I have listed down some pain points in this approach would be very happy to get some feedback -

1. All the components that would require a browser title would need to have an extra `renderTabTitle` prop
2. If the function is kept on the top level, when going to another component without a title, the previous title gets retained instead of just showing `CERN Library Catalogue` again as it should, fix is it set the `document.title =` to the components.

### DIFF
--- a/src/lib/pages/backoffice/Document/DocumentDetails/DocumentDetails.js
+++ b/src/lib/pages/backoffice/Document/DocumentDetails/DocumentDetails.js
@@ -42,6 +42,13 @@ export default class DocumentDetails extends Component {
     if (!samePidFromRouter) {
       fetchDocumentDetails(documentPid);
     }
+
+    const { renderTabTitle, data } = this.props;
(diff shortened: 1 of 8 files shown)


src/lib/pages/backoffice/Document/DocumentDetails/DocumentDetails.js
```diff
@@ -104,8 +111,10 @@ DocumentDetails.propTypes = {
       documentPid: PropTypes.string,
     }),
   }).isRequired,
+  renderTabTitle: PropTypes.func,
```
kpsherva, line 114: if you add as required prop, we can avoid the if statement on each function call

PR #619, opened 2024-04-10
EItem: Accomodate EItem type for facets, details page, cards and editor

sakshamarora1:

### Description

closes: https://github.com/CERNDocumentServer/cds-ils/issues/805
closes: https://github.com/CERNDocumentServer/cds-ils/issues/865

Screenshot 2024-04-10 at 16 10 04
Screenshot 2024-04-10 at 16 10 27
Screenshot 2024-04-10 at 16 10 47
Screenshot 2024-04-10 at 16 11 03

### DIFF
--- a/src/lib/modules/Document/DocumentListEntry/DocumentListEntry.js
+++ b/src/lib/modules/Document/DocumentListEntry/DocumentListEntry.js
@@ -20,6 +20,22 @@ class DocumentListEntry extends Component {
     this.metadata = props.metadata;
   }
 
+  renderEItemTags = (eitems) => {
+    if (eitems.total === 0) {
+      return null;
+    }
+    return [...new Set(eitems.hits.map((eitem) => eitem.eitem_type))].map(
+      (tag) => {
+        return (
+          <List.Item key={tag}>
+            <List.Icon name="desktop" />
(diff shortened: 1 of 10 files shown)


src/lib/modules/Document/DocumentListEntry/DocumentListEntry.js
```diff
@@ -20,6 +20,22 @@ class DocumentListEntry extends Component {
     this.metadata = props.metadata;
   }
 
+  renderEItemTags = (eitems) => {
```
kpsherva, line 23: for better understanding in semantic ui context, and to make the connection with eitem types defined above:
```suggestion
  renderEItemTypeLabels = (eitems) => {
```

src/lib/pages/backoffice/EItem/EItemEditor/schema.js
```diff
@@ -74,7 +84,7 @@ export const schema = () => {
         type: 'array',
       },
     },
-    required: ['document_pid'],
+    required: ['document_pid', 'eitem_type'],
```
kpsherva, line 87: I don't remember if possible to set a default value for required field in the form: in most of the cases it will be an e-book, so we could help the librarians to have it pre-filled

PR #622, opened 2024-04-30
FrontSite: Implement Self Checkout Station using Barcode Scanner

sakshamarora1:

### Description

closes https://github.com/CERNDocumentServer/cds-ils/issues/839


package.json
```diff
@@ -156,5 +156,7 @@
       "enzyme-to-json/serializer"
     ]
   },
-  "dependencies": {}
+  "dependencies": {
+    "html5-qrcode": "^2.3.8"
```
kpsherva, line 160: it should be in peer dependencies, like the rest of the packages

rollup.config.js
```diff
@@ -53,6 +53,7 @@ export default {
       sourcemap: true,
     },
   ],
+  external: ['html5-qrcode'],
```
kpsherva, line 56: just for me to understand, since it is new, why is it needed?

src/lib/pages/frontsite/SelfCheckout/SelfCheckout.js
```diff
@@ -0,0 +1,144 @@
...
+      this.toggleModal();
+    }
+  };
+
+  qrboxFunction = (viewfinderWidth, viewfinderHeight) => {
+    return {
+      width: viewfinderWidth * 0.9,
+      height: viewfinderWidth * 0.3,
+    };
+  };
```
ntarocco, line 52: This function knows basically nothing where or how it is used, I would make it a bit more self-explanatory. You are also not using the Height, is that correct?
```suggestion
  getScannerArea = (cameraWidth, cameraHeight) => {
    return {
      width: cameraWidth * 0.9,  // 90% width
      height: cameraHeight * 0.3,  // 30% height
    };
  };
```

src/lib/pages/frontsite/SelfCheckout/SelfCheckout.js
```diff
@@ -0,0 +1,144 @@
...
+
+  startScanner = (selectedDeviceId, onScanSuccess, onScanFailure) => {
+    const scanner = new Html5Qrcode(selectedDeviceId, {
+      formatsToSupport: [
+        Html5QrcodeSupportedFormats.CODE_128, // config
```
ntarocco, line 57: What does `// config` mean? Please remove it if not needed. Same below.

src/lib/pages/frontsite/SelfCheckout/SelfCheckout.js
```diff
@@ -0,0 +1,144 @@
...
+import { SelfCheckoutModal } from './SelfCheckoutModal';
+import { invenioConfig } from '@config';
+import _get from 'lodash/get';
+
+class SelfCheckout extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      scanner: null,
```
ntarocco, line 17: I think that it does not make much sense to put `scanner` in the state. Why not on `this` instead?

src/lib/components/BarcodeScanner/BarcodeScanner.js
```diff
@@ -0,0 +1,79 @@
...
+
+class BarcodeScanner extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      selectedDeviceId: null,
+    };
+
+    this.scanner = null;
```
ntarocco, line 15: We should not have this here, we are adding a dependency to the parent component. 

The `stopScanner()` should receive no params, as it is a callback to the parent cmp.
In the parent cmp, the stopScanner func should:
```
stopScanner:
  this.stopScanner(this.scanner);
```

src/lib/components/BarcodeScanner/BarcodeScanner.js
```diff
@@ -0,0 +1,79 @@
...
+      const videoInputDevices = await getCameras();
+      if (videoInputDevices.length >= 1) {
+        this.setState({ selectedDeviceId: videoInputDevices[0].id });
+      }
+    } catch (err) {
+      console.error(err);
+    }
+
+    const { selectedDeviceId } = this.state;
+    this.scanner = startScanner(selectedDeviceId, onScanSuccess, onScanFailure);
```
ntarocco, line 40:
```suggestion
    startScanner(selectedDeviceId, onScanSuccess, onScanFailure);
```

src/lib/pages/frontsite/SelfCheckout/SelfCheckout.js
```diff
@@ -0,0 +1,144 @@
...
+import PropTypes from 'prop-types';
+import Overridable from 'react-overridable';
+import { Container, Header } from 'semantic-ui-react';
+import { BarcodeScanner } from '@components/BarcodeScanner';
+import { Html5Qrcode, Html5QrcodeSupportedFormats } from 'html5-qrcode';
+import { SelfCheckoutModal } from './SelfCheckoutModal';
+import { invenioConfig } from '@config';
+import _get from 'lodash/get';
+
+class SelfCheckout extends React.Component {
```
ntarocco, line 12: Actually, ignore my other comments: there is a mix of responsibilities here that we need to untangle.

This cmp should only care about checking out or display messages. We should expect to have only one callback here:
```
onBarcodeScanned(barcode) {
  ... checkout ...
```

And all the rest should be in the cmp that is handing the scanner. Here, both cmps know how to use the scanner, while we really wanted to encapsulate/hide the scanner in the child cmp (or another object).
If you have to pause/resume the camera, you should pass down a prop that says `cameraHidden` `true/false` or `checkingOut` or similar (find better naming), and the child cmp should know that it has to call pause or resume.
Basically, the parent should not know much how to use the camera. It is your child BarcodeScanner that is the glue in-between the React app and the library.

rollup.config.js
sakshamarora1: So this tells Rollup not to bundle this module's code to prevent `THIS_IS_UNDEFINED` error.

src/lib/pages/frontsite/SelfCheckout/SelfCheckout.js
sakshamarora1: Thanks, I will update the function! Also, yes I should use the height.

src/lib/pages/frontsite/SelfCheckout/SelfCheckout.js
sakshamarora1: Thanks a lot for the input, having 2 components share responsibilities isn't ideal, the other aspect I kept in mind was to implement the components such that the library becomes easily replaceable later on if needed. Do you think it'll be better to have 3 components?
- Parent as it is right now, with it not knowing how to access the camera/scanner (it'll encapsulate the UI elements including SelfCheckout) [name: SelfCheckoutPage]
- Middle component which has all the controls/functionality (a generic connector between the app and a scanner library) [name: SelfCheckout]
- Child component which is where the scanner library related code goes) [name: BarcodeScanner]

Do you think that will make it complex?

package.json
```diff
@@ -156,5 +156,7 @@
...
+  }
```
kpsherva, line 161: are the images free use licensed?

src/lib/components/BarcodeScanner/BarcodeScanner.js
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  initializeScanner = async () => {
+    const { startScanner, onScanSuccess, onScanFailure } = this.props;
+
+    this.scanner = startScanner(this.deviceId, onScanSuccess, onScanFailure);
+  };
+
+  render() {
+    return <Container id={this.deviceId} style={{ width: '80%' }} />;
```
kpsherva, line 30: please move the styling to site.overrides

src/lib/pages/frontsite/SelfCheckout/SelfCheckout.js
kpsherva: I agree with Nico on the separation of concerns, lets discuss it in person!

package.json
kpsherva: still not moved to peer deps

src/lib/components/BarcodeScanner/BarcodeScanner.js
```diff
@@ -0,0 +1,77 @@
...
+      /* verbose= */ false
+    );
+  };
+
+  stopScanner = () => {
+    this.scanner.stop();
+  };
+
+  render() {
+    return <Container id={this.deviceId} style={{ width: '80%' }} />; // TODO: Move to overrides
```
kpsherva, line 63: please move to overrides, this TODO will be forgotten

src/lib/config/defaultConfig.js
```diff
@@ -88,6 +88,7 @@ export const APP_CONFIG = {
   HOME_SEARCH_BAR_PLACEHOLDER:
     'Search for books, ebooks, series, journals and standards.',
   PATRON_PROFILE_MAX_RESULTS_SIZE: 100,
+  ENABLE_SELF_CHECKOUT: false,
```
kpsherva, line 91: we are missing a PR setting this to True in cds-ils :)

src/lib/pages/frontsite/SelfCheckout/state/actions.js
```diff
@@ -0,0 +1,79 @@
...
+      dispatch(sendErrorNotification(error));
+    }
+  };
+};
+
+export const checkoutItem = (
+  documentPid,
+  itemPid,
+  patronPid,
+  force = false
```
kpsherva, line 55: force is already false by default in the loanApi , this one is a bit redundant

src/lib/routes/frontsite/frontsiteUrls.js
```diff
@@ -11,6 +11,7 @@ const FrontSiteRoutesList = {
   openingHours: `${FrontSiteBase}opening-hours`,
   seriesDetails: `${FrontSiteBase}series/:seriesPid`,
   errors: `${FrontSiteBase}error`,
+  selfCheckout: `${FrontSiteBase}selfcheckout`,
```
kpsherva, line 14: when configuration is not set to true, we see blank page here instead of 404

src/lib/pages/frontsite/SelfCheckout/state/actions.js
```diff
@@ -0,0 +1,79 @@
...
+) => {
+  return async (dispatch) => {
+    try {
+      const response = await loanApi.doCheckout(
```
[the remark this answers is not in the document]
kpsherva: to summarise the discussion on the permissions:
- we need to modify permissions for the patron to be able to checkout the loan only for their own patron_id (not for other users - simple comparison between transaction_user_pid and patron_pid)
- the above does not apply in terms of the librarian, they should be able to checkout books for anyone
- ticket will be created to address possible bulk scripting issue for malicious intent

src/lib/pages/frontsite/SelfCheckout/state/actions.js
```diff
@@ -22,7 +22,7 @@ export const notifyResultMessage = (message) => {
 
 const searchItem = async (dispatch, term) => {
   const response = await itemApi.list(itemApi.query().withBarcode(term).qs());
-  const item = _first(response.data.hits);
+  const item = _first(response.data.hits) || {};
```
kpsherva, line 25: did you try with null?

PR #623, opened 2024-05-16
frontsite: DocumentSubjects: Don't render if subjects list is empty

sakshamarora1:


src/lib/modules/Document/DocumentSubjectSearch.js
```diff
@@ -25,6 +26,34 @@ import { InvenioRequestSerializer } from 'react-searchkit';
 import { DocumentSubjectGrid } from './DocumentSubjectLayout';
 import { Error as IlsError } from '@components/Error';
 import { SearchControlsMobile } from '@modules/SearchControls/SearchControlsMobile';
+import { ILSParagraphPlaceholder } from '@components/ILSPlaceholder';
+
+export const DocumentSubjects = ({ metadata, isLoading }) => {
+  const renderDocumentSubjectSearch = _get(metadata, 'subjects', null) !== null;
```
kpsherva, line 32:
```suggestion
  const renderDocumentSubjectSearch = _isEmpty(metadata.subjects);
```

kpsherva, line 32: this statement should be used in the parent component, thanks to that we avoid rendering it at all (and allocating memory)

kpsherva (approved):

salomerohr:
great!

PR #624, opened 2024-05-21
frontsite: DocumentDetails: Display multipart above Literature Title

sakshamarora1:

### Description

closes: https://github.com/CERNDocumentServer/cds-ils/issues/734

Screenshot 2024-06-28 at 15 12 40

Screenshot 2024-06-28 at 15 13 10

### DIFF
--- a/src/lib/pages/frontsite/Documents/DocumentDetails/DocumentPanel/DocumentTitle.js
+++ b/src/lib/pages/frontsite/Documents/DocumentDetails/DocumentPanel/DocumentTitle.js
@@ -5,6 +5,8 @@ import React, { Component } from 'react';
 import { Grid, Header, Label } from 'semantic-ui-react';
 import LiteratureTitle from '@modules/Literature/LiteratureTitle';
 import { Media } from '@components/Media';
+import { FrontSiteRoutes } from '@routes/urls';
+import { Link } from 'react-router-dom';
 
 export class DocumentTitle extends Component {
   subtitle() {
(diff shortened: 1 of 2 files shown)


src/lib/pages/frontsite/Documents/DocumentDetails/DocumentPanel/DocumentTitle.js
```diff
@@ -19,12 +21,38 @@ export class DocumentTitle extends Component {
...
+    ) : null;
+  }
+
   render() {
     const { metadata } = this.props;
     const volume = _get(metadata, 'relations.multipart_monograph[0].volume');
 
     return (
       <>
+        <Header>
```
kpsherva, line 51: if you don't check if the volume number and relation metadata exists before rendering the Header component, you will most likely end up with an empty space on top of the column (for regular books without multipart and serials)

src/lib/pages/frontsite/Documents/DocumentDetails/DocumentPanel/DocumentTitle.js
```diff
@@ -19,12 +21,38 @@ export class DocumentTitle extends Component {
...
+
+  relationsTitle(relationsData) {
+    return !_isEmpty(relationsData) ? (
+      <Header.Subheader>
+        {relationsData.map((rel, index) => (
+          <>
+            <Link key={rel.pid_value} to={this.getLinkTo(rel)}>
+              {rel.record_metadata.title}
+            </Link>
+            {index === relationsData.length - 1 ? ' featuring:' : ' and '}
```
kpsherva, line 38: not sure I understand why we need "and"?

sakshamarora1: Because it is a list, there can be multiple relations

kpsherva: normally, a volume can belong only to one multipart monograph (but multiple serials, that's why it is modelled as a list), but for this particular case we should display only one multipart

ntarocco:
The PR looks good, but I have some doubts on the UX/design of the `featuring:` part. To me, is not visible and not very understandable.
Screenshot 2024-05-27 at 09 37 51

src/semantic-ui/site/globals/site.overrides
```diff
@@ -357,6 +357,10 @@ a {
     color: #1aa3ff;
   }
 
+  a.relations-title{
+    color: black;
```
kpsherva, line 361: I would leave it blue, as a link, and have it as cursive instead

PR #626, opened 2024-05-26
EItem: Make E-Book default value in editor

sakshamarora1:


salomerohr:
could we just have E-book instead of E-Book? Thanks.

PR #640, opened 2024-06-28
frontsite: DocumentCirculation: Fix margin for Divider

sakshamarora1:


src/lib/pages/frontsite/Documents/DocumentDetails/DocumentCirculation/DocumentEItems.js
```diff
@@ -67,7 +67,7 @@ export class DocumentEItems extends Component {
             </Button>
           </>
         )}
-        <Divider horizontal fitted>
+        <Divider className="document-circulation" horizontal>
```
kpsherva, line 70: the `mt-0 mb-0` classes should work, no need to add more css

PR #649, opened 2024-10-31
wip

ntarocco:


src/lib/modules/Loan/backoffice/LoanList/LoanListEntry.js
```diff
@@ -79,13 +80,19 @@ export class LoanListEntry extends Component {
...
             <Grid.Column width={2} textAlign="center">
-              {deliveryMethod && (
+              {deliveryMethod ? (
                 <>
                   {delivery}{' '}
                   {deliveryMethod.iconClass && (
                     <Icon className={deliveryMethod.iconClass} />
                   )}
                 </>
+              ) : (
```
kpsherva, line 90: I think there was a specified config for this, can't we use it?

src/lib/pages/backoffice/Loan/LoanDetails/LoanMetadata/LoanMetadata.js
```diff
@@ -34,11 +34,20 @@ export class LoanMetadata extends Component {
     if (delivery) {
       const deliveryMethod =
         invenioConfig.CIRCULATION.deliveryMethods[delivery];
+      const isSelfCheckout = delivery === 'SELF-CHECKOUT';
       return (
         <>
-          {deliveryMethod.text}{' '}
-          {deliveryMethod.iconClass && (
-            <Icon className={deliveryMethod.iconClass} />
+          {isSelfCheckout ? (
```
kpsherva, line 40: same as above

src/lib/pages/frontsite/SelfCheckout/SelfCheckout.js
```diff
@@ -41,50 +38,10 @@ class SelfCheckout extends React.Component {
...
-      if (shouldShowModal) {
-        this.toggleModal(true);
-      } else {
-        this.toggleModal(false);
-      }
+      this.toggleModal(true);
     }
   };
 
-  itemStatus = (item) => ({
```
kpsherva, line 55: looks like a lot of the code deleted here, just to understand, what was it replaced by?

anikachurilova: Errors are handled on the backend now. See here: https://github.com/inveniosoftware/invenio-app-ils/pull/1223

src/lib/modules/Loan/backoffice/LoanList/LoanListEntry.js
anikachurilova: Used the config for the other occurrence of this, but here had to leave this piece of code because the logic of calculating `deliveryMethod` above (see line 21) is that it won't show the delivery method is the state of the loan is not `PENDING`. And for self-checkout we need to show it regardless.

src/lib/pages/frontsite/Documents/DocumentDetails/LoanRequestForm/LoanRequestForm.js
```diff
@@ -24,11 +24,13 @@ class LoanRequestForm extends Component {
       invenioConfig.CIRCULATION.deliveryMethods
     );
     this.deliveryMethods = this.withDeliveryMethod
-      ? Object.keys(invenioConfig.CIRCULATION.deliveryMethods).map((key) => ({
-          key: key,
-          value: key,
-          text: invenioConfig.CIRCULATION.deliveryMethods[key].text,
-        }))
+      ? Object.keys(invenioConfig.CIRCULATION.deliveryMethods)
+          .filter((key) => key !== 'SELF-CHECKOUT')
```
anikachurilova, line 28: Prevents from printing 'Self-checkout' in here:
Screenshot 2024-11-07 at 15 31 07

sakshamarora1: Is this also taken care of in the backend if we do it via the API?

src/lib/config/defaultConfig.js
```diff
@@ -175,6 +175,10 @@ export const RECORDS_CONFIG = {
         text: 'To my office',
         iconClass: 'dolly',
       },
+      'SELF-CHECKOUT': {
```
ntarocco, line 178: Honestly, I think that I would not have added the delivery method here, and a condition to filter it out.
It is a special one. I think that it is safer to have an extra config key, e.g. `deliveryMethodSelfCheckout: ....`

src/lib/pages/frontsite/Documents/DocumentDetails/LoanRequestForm/LoanRequestForm.js
ntarocco: > Is this also taken care of in the backend if we do it via the API?

Not really, it is a bit complicated to handle in the backend. I would not filter it out here, as I would use a different config var.
And for the backend, someone will have to know that SELF-CHECKOUT can be pushed. And... not very harmful, it is just a text (no different behaviour).

sakshamarora1 (approved):

src/lib/pages/frontsite/SelfCheckout/state/actions.js
```diff
@@ -51,20 +43,14 @@ export const selfCheckOutSearch = (term) => {
   };
 };
 
-export const checkoutItem = (documentPid, itemPid, patronPid) => {
+export const selfCheckOut = (documentPid, itemPid, patronPid) => {
```
ntarocco, line 46: The related tests in `actions.test.js` are not updated (we don't use the item anymore),  and they are failing.

PR #654, opened 2025-05-30
backoffice: added NOT-SPECIFIED delivery method

JakobMiesner:
closes: https://github.com/CERNDocumentServer/cds-ils/issues/968

### DIFF
--- a/src/lib/modules/ESSelector/ESSelectorLoanRequest.js
+++ b/src/lib/modules/ESSelector/ESSelectorLoanRequest.js
@@ -35,9 +35,17 @@ export default class ESSelectorLoanRequest extends Component {
           text: invenioConfig.CIRCULATION.deliveryMethods[key].text,
         }))
       : [];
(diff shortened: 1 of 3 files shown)


src/lib/modules/ESSelector/ESSelectorLoanRequest.js
```diff
@@ -36,7 +36,7 @@ export default class ESSelectorLoanRequest extends Component {
         }))
       : [];
     this.state['deliveryMethod'] = this.withDeliveryMethod
-      ? this.deliveryMethods[1].value
+      ? this.deliveryMethods[0].value
```
kpsherva, line 39: I know it was like this in the past, but can you think of any other way of defining the delivery method than accessing the first list element? IMO this way will not be very reliable, order in config might change, especially when someone else configures and uses it

kpsherva (approved):

PR #659, opened 2025-08-26
Improve handling of 'self checkout' action in backoffice loan page

JakobMiesner:
Omit option to self_checkout on loan with no items in backoffice. Previously this produced the following error, as the system does not know which item to 'self checkout'.
Screenshot 2025-08-26 at 11 15 23

Also changed formatting of the 'self_checkout' action button text from capitalize first letter to [start case](https://en.wikipedia.org/wiki/Letter_case#Stylistic_or_specialised_usage)
Screenshot 2025-08-26 at 11 16 49


salomerohr:
@JakobMiesner 
1. We are not sure to understand the error thing. Could you please explain it to us? It also displays the message when an item is available.
2. Librarians won't do 'self-checkout' so the button is misleading in backoffice. To avoid wrong actions, it should be disabled?Librarians are only doing regular check out for patrons.

PR #703, opened 2026-07-16
update(notifications): added links to notifications for mentioned documents

### LINKED ISSUE #833: Error message: direct link to record
In back office. When we get an error message, it would be practical to have a direct link to the concerned document. For example:
error_message

EDIT:
Please check other occurrences of references to IDs,
change the reference of the clickable barcode to lead to the physical item search,
improve the second error message to be more understandable
check if we need any links in the importer errors

callumlewis05:
### Description
* Added generic support for notifications to now have links but left params undefined by default to ensure existing notifications are unaffected. 
* Updated check-in item not on loan/found to have a link to the search for the entered barcode in the backoffice. 
* Left other notifications as they are as none particularly looked like they needed a link to elsewhere but will be a simple future fix if needed.

Error message with link:
image

Link clicked with item not existing (opens in new tab): 
image

Link clicked with item existing (opens in new tab):
image

* closes https://github.com/CERNDocumentServer/cds-ils/issues/833

### DIFF
--- a/src/lib/components/Notifications/Notifications.js
+++ b/src/lib/components/Notifications/Notifications.js
@@ -18,7 +18,7 @@ export default class Notifications extends Component {
         <>
           {before}
           <a
-            style={{ textDecoration: 'underline' }}
+            className="notification-link"
             href={notification.link}
             target="_blank"
             rel="noopener noreferrer"
--- a/src/lib/components/Notifications/Notifications.js
+++ b/src/lib/components/Notifications/Notifications.js
@@ -8,6 +8,27 @@ import PropTypes from 'prop-types';
 import { ErrorMessage, SuccessMessage, WarningMessage } from './messages';
 
 export default class Notifications extends Component {
+  renderMessageContent = (notification) => {
+    const [before, after] = notification.content.split(notification.linkDisplayName);
+
+    if (notification.link) {
+      return (
+        <>
+          {before}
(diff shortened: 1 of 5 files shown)


src/lib/components/Notifications/Notifications.js
```diff
@@ -8,6 +8,29 @@ import PropTypes from 'prop-types';
...
+    const [before, after] = notification.content.split(
+      notification.linkDisplayName
+    );
+
+    if (notification.link && notification.linkDisplayName) {
+      return (
+        <>
+          {before}
+          <a
+            style={{ textDecoration: 'underline' }}
```
kpsherva: please avoid adding inline styling - this is not the best practice. Modify the semantic ui theme files instead

kpsherva (approved):

src/semantic-ui/site/globals/site.overrides
```diff
@@ -485,6 +485,10 @@ a {
     .message {
       padding-right: 2.5em;
     }
+
+    .notification-link {
```
kpsherva: do we need to add a new CSS class or there is already one existing that provides the link styling?
I would prefer not to add CSS maintenance

callumlewis05: I have not been able to find an already existing one

PR #713, opened 2026-08-03
edit(loans): loan dates now only editable when active or pending

### LINKED ISSUE #1004: Editable loan dates on 'returned' loan should not be possible
**Describe the bug**
It looks we can edit loan dates on returned loans.

**To Reproduce**
On the loan -> Edit loan dates
It should only be possible for ongoing loans.

**Expected behavior**
It should not be possible if the loan is 'returned', e.g. we should have a message saying the loan is finished and therefore the dates cannot be edited. Suggest that a new loan should be issued.

callumlewis05:
Loans dates can now only be edited if the loan is active or pending. When the edit button is clicked a popup appears with an error message if the loan is returned or cancelled.

UI after clicking the edit dates return button for each type of loan:

Pending loan:
image

Cancelled loan:
image

Item on loan (active):
image

Returned:
image

* closes https://github.com/CERNDocumentServer/cds-ils/issues/1004

### DIFF
--- a/src/lib/pages/backoffice/Loan/LoanDetails/LoanUpdateDates/LoanUpdateDates.js
+++ b/src/lib/pages/backoffice/Loan/LoanDetails/LoanUpdateDates/LoanUpdateDates.js
@@ -35,29 +35,33 @@ export default class LoanUpdateDates extends Component {
     return metadata.state === 'CANCELLED';
   };
 
-  isActiveOrCompleted = () => {
+  isReturned = () => {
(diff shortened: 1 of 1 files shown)


kpsherva (commented):
LGTM, could you add a screenshot of changes? 
Please try to provide screenshots any time there is a change in the interface, it makes it easier to evaluate in terms of the UX.

kpsherva (approved):

PR #715, opened 2026-08-07
add(tab titles): added unique tab titles for each page in frontoffice

### LINKED ISSUE #793: Display title of page in Tab
The tabs are displaying 'CERN Library Catalogue' and it can be quite difficult to differentiate them when we work on several tabs opened at the same time.

On CDS or INSPIRE, the title of the page is displayed:
tabs

Best regards,

Salomé

callumlewis05:
Gave each page in the front and back office a unique title to help easily identify them whilst many tabs are open.

image

* closes https://github.com/CERNDocumentServer/cds-ils/issues/793

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,6 +66,9 @@
     "@rollup/plugin-commonjs": "^12.0.0",
     "@rollup/plugin-node-resolve": "^8.0.1",
     "@types/jest": "^29.5.0",
+    "@visx/responsive": "^3.12.0",
+    "@visx/scale": "^3.12.0",
+    "@visx/shape": "^3.12.0",
     "axios": "^1.8.2",
     "axios-mock-adapter": "^2.0.0",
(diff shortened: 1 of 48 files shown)


package.json
```diff
@@ -157,5 +157,8 @@
   },
   "engines": {
     "node": ">=22.0.0"
+  },
+  "dependencies": {
+    "react-helmet-async": "^3.0.0"
```
kpsherva: for this library - did you consider any other alternatives? Could you summarise? It is slightly concerning that the library has a single contributor, plus it is already forked from elsewhere, I am worried about future maintainability

callumlewis05: At the time I looked at other ways of doing it and I found there is a way of doing it in plain JS but I remember it was more complex and I believe there were other suitable libraries but this was the best with the most recent updates and most used. If you think it is worth it for one less dependency I can change it to plain JS?

PR #718, opened 2026-08-24
update(literature): add note saying office delivery is free

### LINKED ISSUE #1106: add the sentence "(service free of charge)" next to the option "Delivery options > To my office"
Image

## Context 

Relevant ticket: RQF3881260

- This would be extremely helpful for us, as we have recently received a significant number of questions and encountered some misunderstandings from patrons who assumed that this delivery service implies a fee. Adding this clarification directly next to the option would help avoid confusion and save us time.

callumlewis05:
Add note stating delivery to an office is free

* closes https://github.com/CERNDocumentServer/cds-ils/issues/1106

### DIFF
--- a/src/lib/config/defaultConfig.js
+++ b/src/lib/config/defaultConfig.js
@@ -181,7 +181,7 @@ export const RECORDS_CONFIG = {
         iconClass: 'warehouse',
       },
       DELIVERY: {
-        text: 'To my office',
+        text: 'To my office (free delivery service)',
         iconClass: 'dolly',
       },
     },


src/lib/config/defaultConfig.js
```diff
@@ -181,7 +181,7 @@ export const RECORDS_CONFIG = {
         iconClass: 'warehouse',
       },
       DELIVERY: {
-        text: 'To my office',
+        text: 'To my office (free delivery service)',
```
kpsherva: unfortunately we can't assume this for any ILS - this should be implemented only for CDS-ILS

PR #719, opened 2026-08-25
upgrade(selfcheckout): added link and page refresh on redirect

### LINKED ISSUE #978: Self-checkout improvement
Original request: https://cern.service-now.com/nav_to.do?uri=u_request_fulfillment.do?sysparm_query=number=RQF3055743

Could you please explore if it would be possible, when an item with status 'missing' is scanned via the self-checkout, that the item automatically becomes back to 'can circulate' in our backoffice (as the copy is there) and that the loan is issued for the user?

callumlewis05:
Updates profile page with loans by updating component once selfcheckout has finished and added link to selfcheckout to make it more accessible than just QR code in library.

* closes https://github.com/CERNDocumentServer/cds-ils/issues/978

### DIFF
--- a/src/lib/components/ILSMenu/ILSMenu.js
+++ b/src/lib/components/ILSMenu/ILSMenu.js
@@ -44,6 +44,10 @@ class ILSMenu extends Component {
           </>
         ) : null}
         <Dropdown.Divider />
+        <Dropdown.Item as={Link} to={FrontSiteRoutes.selfCheckout}>
+          Self-checkout
+        </Dropdown.Item>
+        <Dropdown.Divider />
         <Dropdown.Item onClick={this.logout}>Sign out</Dropdown.Item>
       </Dropdown.Menu>
     );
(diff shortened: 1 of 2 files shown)


src/lib/components/ILSMenu/ILSMenu.js
```diff
@@ -44,6 +44,10 @@ class ILSMenu extends Component {
           </>
         ) : null}
         <Dropdown.Divider />
+        <Dropdown.Item as={Link} to={FrontSiteRoutes.selfCheckout}>
+          Self-Checkout
```
kpsherva: could you make sure first that self checkout is available for whole ILS not only CDS-ILS (I mean as a feature). If it is only CDS-ILS we can't add it for any instance

callumlewis05: I have checked and there are 20+ other files related to self checkout in the react repo so it is in ILS as a whole not just CDS-ILS

